PR #37, opened 2017-10-22
Fix DateTime.fromJSDate input handling

dstelljes:
Restrict input to Dates; produce invalid DateTimes from invalid Dates.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -17,6 +17,7 @@ import {
 } from './errors';
 
 const INVALID = 'Invalid DateTime',
+  INVALID_INPUT = 'invalid input',
   UNSUPPORTED_ZONE = 'unsupported zone',
   UNPARSABLE = 'unparsable';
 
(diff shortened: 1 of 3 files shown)


dstelljes:
Thoughts on the lint? `isNaN` triggers no-restricted-globals, `Number.isNaN` isn't safe, `typeof o === 'number' && o !== o` triggers no-self-compare.

icambron:
Hmm, I've been using `Number.isNaN`. Why isn't it safe? My understanding was that it was *more* safe than the `isNaN` global and hence its restrictions. But I'm just getting that from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN).

icambron:
The patch looks good, btw

dstelljes:
Safe from a browser compatibility standpoint, should have been more clear. I was thinking it was a newer thing, but it looks like it's just IE that doesn't support it.

icambron:
Ah, I see. I polyfilled it here: https://github.com/icambron/luxon/blob/master/src/luxon.js#L4

icambron:
And thanks for the contribution!

dstelljes:
D'oh, didn't even check. And thanks for the library, it's a really neat project!

PR #38, opened 2017-10-22
Make prettier run on precommit hooks automatically

JemBijoux:
Normal commit hooks often require modifying files within the .git
directory manually. However, `husky` and `pre-commit` together make it
so you don't have to do anything special to run prettier against files
automatically as they are being committed. This ensures that any file
changed is automatically updated when committed.

- Add husky and precommit to the package, along with a precommit script
that will run prettier for any staged file

Note that this is the process that prettier [recommends here](https://github.com/prettier/prettier#pre-commit-hook). This addresses https://github.com/icambron/luxon/issues/27

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,13 +3,16 @@
   "version": "0.0.5",
   "description": "Immutable date wrapper",
   "author": "Isaac Cambron",
-  "keywords": [
-    "date",
-    "immutable"
-  ],
+  "keywords": ["date", "immutable"],
   "repository": "https://github.com/icambron/luxon",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


JemBijoux:
Hey Isaac. So I was chatting with Maggie Pint during node interactive about maybe helping out a bit with something `moment` related. She mentioned that I should check out what you've got here with Luxon though and offer to help here instead. So here I am!

icambron:
Thanks for the help, @JemBijoux. This is neat; I hadn't known about Husky. I'm testing it out.

PR #43, opened 2017-10-29
Don’t publish unused files to npm.

tobli:
Only include the generated cjs file in the npm package (exclude sites, site, src etc.). This removed tens of megabytes from the npm package. Note that package.json, changelog.md, license.txt and readme.md are not listed, since they are always included by npm publish.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,5 +57,6 @@
   "engines": {
     "node": "*"
   },
+  "files": ["build/cjs/luxon.js"],
   "license": "MIT"
 }


icambron:
Hmm, right. I definitely agree we don't want `node_modules` in there. For some reason I thought those were automatically excluded. But do we need the `/src` dir? I have some vague recollection that some toolchains would want to consume that directly.

In the meantime, I'm going to merge this. Thanks

Issue #48, opened 2017-10-29
Add convenience getters for month and weekday names


icambron:
These are analogous to `offsetNameShort` and `offsetNameLong`:

 * `dateTime.weekdayShort`
 * `dateTime.weekdayLong`
 * `dateTime.monthShort`
 * `dateTime.monthLong`

mslooten:
You mean like a convenience wrapper for `DateTime.local().toLocaleString({ weekday: 'short' })` or `DateTime.local().toFormat('ccc')` (and the likes)? I'm happy to take a stab at this.

icambron:
@mslooten Yup, that's right. Though I think it might be better implemented as `Info.monthsShort[this.month - 1]`

icambron:
Fixed in #49

PR #49, opened 2017-10-31
Adds weekday & month convenience getters

mslooten:
Adds the convenience getters as discussed in #48

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1,6 +1,7 @@
 import { Duration } from './duration';
 import { Interval } from './interval';
 import { Settings } from './settings';
+import { Info } from './info';
 import { Formatter } from './impl/formatter';
 import { FixedOffsetZone } from './zones/fixedOffsetZone';
 import { LocalZone } from './zones/localZone';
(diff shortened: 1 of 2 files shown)


icambron:
This looks great. I'll figure out why the CI failed

mslooten:
@icambron Thanks for checking. Yeah I was wondering about that too, only thing I'm unsure of is the circular import (datetime.js imports info.js which also imports datetime.js) to be able to use Info. But local tests & builds are fine.

icambron:
Circular imports are OK with Rollup. If the library got bigger it would be a maintenance issue and I'd have to sort it out, but as it is I'm not too worried about it.

icambron:
The CI problem was just Travis having some issues.

Can you add tests for a non-en-US locale? I usually use "fr". The reason for this is that Luxon has some fallback logic for older browsers that treats "en" and "en-US" specially.

icambron:
Looks good, thanks! I'm going to push a new version in the next couple of days that includes this.

PR #50, opened 2017-10-31
Add support for ISO week duration

gbirke:
The `PnW` format for durations was not supported.

### DIFF
--- a/src/impl/regexParser.js
+++ b/src/impl/regexParser.js
@@ -79,14 +79,15 @@ function extractISOTime(match, cursor) {
 
 // ISO duration parsing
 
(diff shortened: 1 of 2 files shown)


icambron:
I think this is a good feature. I want it implemented a little differently, though. Luxon Durations can track weeks directly; it's just that none of the existing parsers return them. In other words, instead of this:

```js
Duration.parseISO("P2W").toObject() // => { days: 14 }
```

It should be:

```js
Duration.parseISO("P2W").toObject() // => { weeks: 2 }
```

I think this actually simplifies things: just assign `week: parseInt(weekStr)` and call it day

gbirke:
Changed the parser code, it really got better, thanks.

What do you think of adding a week test to `math.test.js` to check the week support? Any other places where week support might need to be tested?

icambron:
Yup, `plus`, `minus`, `startOf`, `endOf` should all be tested with weeks, along with `until`, `hasSame`, and `diff`, and `Interval#toDuration`. 

But I'll go ahead and merge this. Thanks!

PR #69, opened 2017-11-19
Support 1-9 digits of fractional seconds parsing ISO times

### LINKED ISSUE #65: Request: Allow a variable number of digits for fractions of a second in DateTime.fromISO
I'm having an issue where I'm using Go to encode my JSON and it outputs dates in ISO format like `"2017-12-02T05:26:01.302911182Z"` with 9 decimal places but your parser will only allow exactly 3. According to the wikipedia page on ISO8601:

> There is no limit on the number of decimal places for the decimal fraction. However, the number of decimal places needs to be agreed to by the communicating parties. For example, in Microsoft SQL Server, the precision of a decimal fraction is 3, i.e., "yyyy-mm-ddThh:mm:ss[.mmm]".

That being said a generic parser like luxon would be more useful if it were more flexible in its ability to parse ISO-formatted dates. I'm curious if you would be open to adjusting the regex to allow for 1-9 digits.

Here's a version that works for my use case:
```js
const isoTimeRegex = /(?:T(\d\d)(?::?(\d\d)(?::?(\d\d)(?:[.,](\d{1,9}))?)?)?(?:(Z)|([+-]\d\d)(?::?(\d\d))?)?)?$/
```

Happy to submit a PR if that helps.

snocorp:
Fixes #65 

This PR will convert the fractional part of the time to nanoseconds by padding the number to 9 digits and then converting to milliseconds by dividing by 1,000,000 and rounding the result.

### DIFF
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -64,6 +64,10 @@ export class Util {
     return ('0'.repeat(n) + input).slice(-n);
   }
 
+  static padEnd(input, n = 9) {
+    return (input + '0'.repeat(n)).slice(0, n);
+  }
+
   static towardZero(input) {
     return input < 0 ? Math.ceil(input) : Math.floor(input);
   }
--- a/src/impl/regexParser.js
+++ b/src/impl/regexParser.js
@@ -67,12 +67,11 @@ function extractISOTime(match, cursor) {
   const local = !match[cursor + 4] && !match[cursor + 5],
     fullOffset = Util.signedOffset(match[cursor + 5], match[cursor + 6]),
     nanosecond = Util.padEnd(match[cursor + 3] || '0'),
-    millisecond = Math.round(parseInt(nanosecond) / 1000000),
     item = {
       hour: parseInt(match[cursor]) || 0,
       minute: parseInt(match[cursor + 1]) || 0,
       second: parseInt(match[cursor + 2]) || 0,
-      millisecond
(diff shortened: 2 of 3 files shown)


src/impl/util.js
```diff
@@ -64,6 +64,10 @@ export class Util {
     return ('0'.repeat(n) + input).slice(-n);
   }
 
+  static padEnd(input, n = 9) {
+    return (input + '0'.repeat(n)).slice(0, n);
```
icambron: This might be better as

```js
input.length > n ? input : input + '0'.repeat(n - input.length)
```

icambron:
This looks good, just the one comment.

src/impl/util.js
icambron: Oh, I see. You were just using the same idiom as `pad` just above it. Nevermind, just leave it :)

icambron:
Thanks!

src/impl/util.js
snocorp: Perhaps `padEnd` is the wrong name for this. The intent of this is similar to `pad` above in that if there were more than `n` characters, I want only the first `n` characters. `pad('123')` above will return `'23'` and `padEnd('123', 2)` should return `'12'`.

icambron: As long as we're consistent. I renamed `pad` to `padStart`. It's possible they should both have clearer names but it seems OK for now.

PR #75, opened 2017-11-20
test: use bitnami prod image version for smaller docker test env

tuananh:
build and then test like normally

    docker build -t icambron/luxon docker
    ./docker/gulp test

old: 292 MB (compressed)
new: 150MB (uncompressed), I haven't pushed to registry to see the size after compressed. an educated guess would be around 50-60MB

I also tried with `node:8-alpine` (official image) but there were a couple of failed tests. like this one.
I'm not sure if it's related to `alpine` or `bitnami:node` (based on debian).
Using `node:8-alpine` would result in even smaller image size (100MB uncompressed) but i want to clarify with you first if those tests are related.

```
FAIL  test/datetime/diff.test.js
  ● DateTime#diff is precise for lower order units

    expect(received).toEqual(expected)

    Expected value to equal:
      {"hours": 2999}
    Received:
      {"hours": 3000}

    Difference:

    - Expected
    + Received

      Object {
    -   "hours": 2999,
    +   "hours": 3000,
      }

      at Object.<anonymous> (test/datetime/diff.test.js:143:5)
          at new Promise (<anonymous>)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.9.0
+FROM bitnami/node:8.9.1-r0-prod
 
 RUN npm install full-icu
 RUN npm install -g gulp-cli


icambron:
That test failure is very confusing. I have no idea why that would happen. Would take some debugging.

I'll pull the one you posted down tomorrow, try it out, and then upload it to the registry.

icambron:
I looked into the Alpine build and it's just broken:

```js
> new Intl.DateTimeFormat('en').resolvedOptions()
{ locale: 'en',
  numberingSystem: 'latn',
  calendar: 'gregory',
  timeZone: 'America/New_York',
  year: 'numeric',
  month: 'numeric',
  day: 'numeric' }
> new Date().getTimezoneOffset()
0
```

In other words, in some ways it acts like it's in the zone specified in the env and in other ways it totally ignores that. That confusion breaks all the tests. I suspect what's happening is that Node understands the zone but the OS does not:

```
$ docker run icambron/luxon date
Mon Nov 20 21:14:39 GMT 2017
```

Where that should have used the TZ env var set in Docker.

But I did test the Bitnami one and it looks fine, so I'm going to merge this as-is and deploy it.

icambron:
Thanks!

PR #78, opened 2017-11-20
fix: typo - missing 'wait'

crhistianr:


crhistianr:
Now that I read it again that might've been intentionally left out

icambron:
It was, but it is poor writing, so I'll reprhase

icambron:
Fixed in af92c70

Issue #79, opened 2017-11-20
Parse/format SQL dates, times, and datetimes


icambron:
Luxon should support SQL times with a new static method called `fromSQL` and instances methods `toSQL`, `toSQLDate`, and `toSQLTime`:

 * `fromSQL` accepts strings like '2016-05-14', '04:12:00.123', and ‘2016-05-14 10:23:54.234’ and should be implemented in the regex parser (see the ISO equivalent method)
 * `toSQLDate`, `toSQLTime`, and `toSQL` should output strings like '2016-05-14', '04:12:00.123', '2016-05-14 10:23:54.234`, respectively. There are analogous ISO-outputting methods to refer to.
 * Times and Datetimes should output up to milliseconds but parse up to nanoseconds

tuananh:
@icambron i'll bite. can you help review it https://github.com/moment/luxon/pull/83

dobromyslov:
What about fromNanoseconds()?

For exampre Google Firestore can return timestamp only in seconds and nanoseconds.

icambron:
No, two reasons we won't do that:

 * It would imply to many people that we support nanosecond precision, which we do not

 * It's easy enough to multiply by 1e6. We always need to trade-off usefulness vs bloat. I don't think that makes the cut

benj5378:
Please add this to the documentation

PR #83, opened 2017-11-21
feat: add fromSQL

tuananh:
- attempt to fix https://github.com/moment/luxon/issues/79
- tests included

### DIFF
--- a/src/impl/regexParser.js
+++ b/src/impl/regexParser.js
@@ -111,14 +111,29 @@ const obsOffsets = {
   PST: -8 * 60
 };
 
-function fromStrings(weekdayStr, yearStr, monthStr, dayStr, hourStr, minuteStr, secondStr) {
+function parseSecondFraction(fraction) {
+  const f = parseFloat('0.' + fraction) * 1000;
+  return Math.ceil(f);
+}
+
+function fromStrings(
+  weekdayStr,
+  yearStr,
+  monthStr,
+  dayStr,
+  hourStr,
+  minuteStr,
+  secondStr,
+  fractionStr
+) {
   const result = {
     year: yearStr.length === 2 ? Util.untruncateYear(parseInt(yearStr)) : parseInt(yearStr),
     month:
       monthStr.length === 2 ? parseInt(monthStr, 10) : English.monthsShort.indexOf(monthStr) + 1,
     day: parseInt(dayStr),
     hour: parseInt(hourStr),
-    minute: parseInt(minuteStr)
+    minute: parseInt(minuteStr),
+    millisecond: fractionStr ? parseSecondFraction(fractionStr) : 0
   };
 
   if (secondStr) result.second = parseInt(secondStr);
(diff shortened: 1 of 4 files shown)


src/impl/regexParser.js
```diff
@@ -189,6 +190,34 @@ function extractASCII(match) {
...
+  return [result, FixedOffsetZone.utcInstance];
+}
+
+function extractSqlTime(match) {
+  const now = new Date(),
+    yearStr = now.getUTCFullYear().toString(),
+    monthStr = (now.getUTCMonth() + 1).toString(),
+    dayStr = now.getUTCDate().toString(),
+    [, hourStr, minuteStr, secondStr] = match,
+    result = fromStrings(0, yearStr, monthStr, dayStr, hourStr, minuteStr, secondStr);
```
icambron: For this one, instead of defaulting the year, month, and date, don't use `fromStrings()` and just return the object with parsed hour, minute, and second. See `fromISOTime` for that. `fromObject` in the main DateTime file will handle defaulting those higher-order units.

src/impl/regexParser.js
```diff
@@ -114,7 +114,8 @@ const obsOffsets = {
 function fromStrings(weekdayStr, yearStr, monthStr, dayStr, hourStr, minuteStr, secondStr) {
   const result = {
     year: yearStr.length === 2 ? Util.untruncateYear(parseInt(yearStr)) : parseInt(yearStr),
-    month: English.monthsShort.indexOf(monthStr) + 1,
+    month:
+      monthStr.length === 2 ? parseInt(monthStr, 10) : English.monthsShort.indexOf(monthStr) + 1,
```
icambron: 👍

src/impl/regexParser.js
```diff
@@ -189,6 +190,34 @@ function extractASCII(match) {
   return [result, FixedOffsetZone.utcInstance];
 }
 
+const sqlDateRegex = /^(\d{4})-(\d\d)-(\d\d)$/,
+  sqlTimeRegex = /^(\d|\d\d):(\d\d):(\d\d)\.(\d\d\d)$/,
+  sqlDateTimeRegex = /^(\d{4})-(\d\d)-(\d\d) (\d|\d\d):(\d\d):(\d\d)\.(\d\d\d)$/;
```
icambron: I strongly suspect there can be an arbitrary number of subsecond digits. I just accepted a fix for that in the ISO analog; you can probably extract the pad, parse, and rounding parts from there and put them in a `parseSecondFraction` function and call that from the handlers. Then just change `\d\d\d` to `\d{1,9}`

icambron:
This is looking pretty good. I left a couple comments

src/impl/regexParser.js
tuananh: you mean fraction second can be `12300000` and i have to trim the `0` right pad to get it?

src/impl/regexParser.js
```diff
@@ -189,6 +190,34 @@ function extractASCII(match) {
...
+    yearStr = now.getUTCFullYear().toString(),
+    monthStr = (now.getUTCMonth() + 1).toString(),
+    dayStr = now.getUTCDate().toString(),
+    [, hourStr, minuteStr, secondStr] = match,
+    result = fromStrings(0, yearStr, monthStr, dayStr, hourStr, minuteStr, secondStr);
+
+  return [result, FixedOffsetZone.utcInstance];
+}
+
+function extractSqlDateTime(match) {
```
icambron: You're not having these things take a cursor argument. But if you did have `extractSQLTime` and `extractSQLDate` use `slice` with a cursor argument, you could actually use `combineExtractors` instead of this function. If that's too unclear, don't worry about it; it's my fault for making this code too difficult to understand and this seems fine.

tuananh: yeah i want to use that too, but from what i understand,  it will combine without the empty space i need in the datetime regex. I'm not sure if my understanding is correct here hence i did not use it.

`const full = regexes.reduce((f, r) => f + r.source, '');`

src/impl/regexParser.js
icambron: Well, more importantly it could be `234567890`, which we should parse as `235`. It's the equivalent of these tests: https://github.com/moment/luxon/blob/master/test/datetime/regexParse.test.js#L259

src/impl/regexParser.js
icambron: Yeah, you need to use a separate regex because of the space. But you don't need to use a separate extractor because you don't capture the space in a capture group. Extractors only care about what units correspond to what positions in the match array.

Now that I think about it, that actually makes me wonder if you can actually just reuse the ISO extractors as-is? You need separate regexes, but aren't the units in the same order?

tuananh:
I fixed those that you comments with exception for `combineRegexes`. i dont quite fully understand it yet.

icambron:
I'm going to go ahead and merge this and maybe later I'll take a crack at doing what I proposed. Thanks for the contribution!

icambron:
See 524c701. I made these changes:

 * Unhardcode UTC. I missed this in the review. I think the way it works is that SQL dates and times without any offset information can be treated however the caller wants. The Luxon way of doing that is to let the caller specify the zone and default to interpreting the input as a local if nothing is specified
 * Looking around, it looks like most (all?) SQL dialects support specifying an offset in the string (like "+06:00") at the end so I added support for that
 * I made the fractional seconds optional, e.g. "10:05:36" is ok
 * I did the thing I mentioned about using `combineExtractors` and reused the iso extractor for the parsed sql times.

Next, I'm going to do #85, where I think I can simplify all the regexes by making `combineRegexes` smarter.

tuananh:
Cool. It looks much cleaner now 👍

PR #98, opened 2017-11-24
Feature: module

marcobiedermann:
Add module entry to package.json.
This way webpack or rollup can use the ES6 module instead of the transpiled ES5 version from babel which includes a lot of shimed code.
If module import is supported the `src/luxon.js` file gets imported otherwise it falls back to `build/node/index.js` when using `require`.
This will save quite a few bytes in the final bundle.

See: https://github.com/rollup/rollup/wiki/pkg.module

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,6 @@
   "engines": {
     "node": "*"
   },
-  "files": ["build/node/luxon.js", "build/cjs-browser/luxon.js"],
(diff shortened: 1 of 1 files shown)


icambron:
Sure. Will release this in 0.2.1

icambron:
And thanks!

marcobiedermann:
@icambron 
Cool! You are very welcome 😄

PR #105, opened 2017-12-01
Add flowtype support and proof-of-concept

CAYdenberg:
This adds support for flow types: the build step works correctly and flow types are linted via ESLint.

Fully annotated `english.js` as a proof-of-concept. The major chore will be to go through the project, but this lays the groundwork as far as getting the system working.

One note: types are stripped using babel-preset-flow ffor linting/testing and using rollup-plugin-flow for the actual build.

### DIFF
--- a/src/impl/english.js
+++ b/src/impl/english.js
@@ -12,7 +12,7 @@ function stringify(obj: Object) {
  */
 
 export class English {
-  static get monthsLong() {
+  static get monthsLong(): Array<string> {
     return [
       'January',
       'February',
@@ -29,15 +29,15 @@ export class English {
     ];
   }
 
-  static get monthsShort() {
+  static get monthsShort(): Array<string> {
     return ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
   }
 
-  static get monthsNarrow() {
+  static get monthsNarrow(): Array<string> {
     return ['J', 'F', 'M', 'A', 'M', 'J', 'J', 'A', 'S', 'O', 'N', 'D'];
   }
 
-  static months(length) {
+  static months(length: string): Array<string> {
     switch (length) {
       case 'narrow':
         return English.monthsNarrow;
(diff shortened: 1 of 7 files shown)


icambron:
Awesome start.

You might also be interested in my old aborted attempt at this, since this does retread some of that ground. See the `flow` branch [here](https://github.com/moment/luxon/tree/flow). It's not only way behind master but it's also a WIP that doesn't work, but maybe there's something worth scavenging. Finally, this will help with the CI error: https://github.com/moment/luxon/blob/flow/gulpfile.js#L170

src/impl/english.js
```diff
@@ -73,58 +74,56 @@ export class English {
...
       case 'long':
-        return English.erasLong;
       default:
-        return null;
+        return English.erasLong;
     }
   }
 
-  static meridiemForDateTime(dt) {
+  static meridiemForDateTime(dt: Object) {
```
ulrikstrid: `Object` is as typesafe as `any`. Could you use a `*` or create a interface that could be built on later?

CAYdenberg: I think the plan would be create a datetime type somewhere and import it. My intention with the PR was to try to get the ball rolling ... I didn't necessarily want to create definitions, just get the pipeline working so that adding defs could be a community effort. If the maintainer(s) feel differently I can of course expand on this ...

icambron: I'm good with getting something simple in and tightening it down later. My worry with this as it stands now is that there's not enough of it to make a foothold. I don't want there to be like a random scattering of types. So I'm hoping someone will take this and add a bunch of stuff in before I do any merging.

ulrikstrid: Do you want people to add types to this PR then?
I could probably spend some time on it although I actually want it in a typescript project. Or maybe this is a good reason to ditch TS and use flow instead 🤔

icambron: I think adding types to this PR probably makes sense.

CAYdenberg: I'll keep plugging away at it when I have a chance, but unfortunately got the flu last week and will be in catchup mode for a while. If other people want to add to it or re-fork it it's totally fine with me.

jedwards1211: @ulrikstrid I remember the author of the React Material UI project said one of the things he learned from writing it is that a lot of people are moving from Flow to TypeScript.  I'm not 100% sure why but I get the impression TypeScript is probably more stable and has more convenient error messages (though I'm a Flow user and have never used TS myself).  So definitely don't hurry the decision to switch to Flow.

jedwards1211:
By the way guys, **please add the flow types in `flow-typed` instead of in this package**, because in `flow-typed` is currently the only way to support multiple versions of Flow (I have a PR to `flow-typed` for a way to put defs for multiple versions of flow within packages themselves, but it won't get merged until `flow-typed` 3.0.0).

icambron:
@jedwards1211 Adding an external types file to this repository is the one thing that definitely will NOT happen :)

src/impl/english.js
ulrikstrid: I have used both and I love the number of typed packages for TypeScript but Flow feels so much better. If there were a larger community around Flow where more typings for packages I would love to use that always.
Also as a consultant I sometimes have to make the "safe bet" and that is still TS these days.

jedwards1211: Huh, what feels better about it?
As far as staying on topic, I'm going to try to PR typedefs for `luxon` that I've been using locally to `flow-typed` today.

jedwards1211: @ulrikstrid btw one very cool thing you may want to check out, which I don't think is currently possible with TS, is [flow-runtime](https://github.com/codemix/flow-runtime).

jedwards1211:
https://github.com/flowtype/flow-typed/pull/1654

icambron:
It looks like the external, community-maintained types have won out. I'm still down with using Flow everywhere, but I'm going to close this in the meantime.

jedwards1211:
Using flow types like in this PR would be more for internal QA than public typedefs anyway

PR #106, opened 2017-12-01
fixed the wrong link to DateTime instead of Duration

lacivert:


icambron:
Because of your other PR this will happen automatically when I rebuild the docs.

lacivert:
👍

PR #107, opened 2017-12-01
fixed for correct documentation

lacivert:
if you don't put the right class here, document generator generates a wrong link to another class (DateTime class)


icambron:
Ah, thanks

PR #111, opened 2017-12-04
Add note about Moment->Luxon format tokens

davidsharp:
The format tokens differ between Moment and Luxon, such that the same format string cannot be used between the two.

This has been added as a note, as it's something that needs to be addressed by users migrating from one to the other.

(This may also deserve a place in the formatting and parsing pages of the docs, if they have Moment users in mind)


icambron:
Sure, thanks

PR #119, opened 2017-12-08
FIx toISODate example

jxjj:
Luxon is looking great! Thanks for your work on this. 

Think the `toISODate()` example should show that it returns a date, not a time.


icambron:
Oops, that's what I get for copying and pasting

icambron:
Thanks! I'll push this to the website soon

PR #129, opened 2017-12-14
fix type of zone in JSDoc for several constructors

jedwards1211:
It was `boolean`, should be `string|Zone` as far as I can tell


icambron:
Ah, thanks!

PR #130, opened 2017-12-14
fix return type for toJSDate

jedwards1211:
not sure why it was `object`.  Seems obvious that it should be `Date`.


icambron:
It's because I copy and paste too much. Thanks!

PR #132, opened 2017-12-19
Update formatting.md

austindebruyn:
Quick typo fix. The line "padded to 2" is duplicated, but the documented difference between `format(H)` and `format(HH)` should be that H is not padded.


icambron:
Thanks

PR #133, opened 2017-12-19
docs: jsdoc interrogation spelling

dvlsg:
Just a quick spelling update for `interrogation`, noticed on the [interval docs](https://moment.github.io/luxon/docs/class/src/interval.js~Interval.html).


icambron:
Thanks

Issue #116, opened 2017-12-06
More robust performance testing


icambron:
See the comments at the top of: https://github.com/moment/luxon/blob/master/benchmarks/benchmark.js

mannanali413:
@icambron i can take this up.

mannanali413:
@icambron please have a look at the above PR. I hope it addresses the requirements you had listed earlier. Presently I have added the benchmark tests only for DateTime . 
Should I create a separate issue for the benchmark tests for other modules and raise another PR addressing the same?

icambron:
Fixed in 8445647

@mannanali413 It's always fine for people to add more benchmarks, but I don't think there needs to be an explicit issue.

PR #137, opened 2017-12-27
benchmark: Run Benchmark2.x tests via Gulp Task

mannanali413:
This pull request allows us to run benchmarks in Benchmark2.x as a Gulp Task . 
The Benchmarks can be written in ES6 . 
Added Gulp tasks for babel transpile of the benchmark files and also executing the benchmarks.

Fixes: https://github.com/moment/luxon/issues/116


icambron:
This looks great! I'm going to play around with it, but I think this is exactly what I was looking for.

icambron:
I tweaked this a bit and then merged it in 8445647. Thanks for the contribution!

PR #138, opened 2017-12-27
fix #124: return undefined from min/max if called with no argument

### LINKED ISSUE #124: `DateTime.{min,max}` throw errors if given no dates
I presume they should return `undefined` instead of throwing exceptions?

jnizet:


icambron:
Awesome. Thanks!

PR #146, opened 2018-01-08
Update Interval.intersection API documentation

jubis:
Add mention about the null value when intersection is empty.
The null may not be the best solution (causes null pointers by surprise) but at least it should be mentioned.


icambron:
Thanks!

PR #148, opened 2018-01-08
Fix documentation of Info.features()

patrickjuchli:
- The property describing timezone support is `zones`, not `timezones`.
- The example uses `feature` instead of `features`


icambron:
Thanks!

PR #159, opened 2018-01-12
Don't use Arrow Function

abouthiroppy:
Change `this` scope.

```
console.log('Fastest is ' + undefined.filter('fastest').map('name'));
                                        ^

TypeError: Cannot read property 'filter' of undefined
```


icambron:
Oops, I must have made that change at the last minute and not actually tested it. Thanks.

PR #160, opened 2018-01-12
Add .editorconfig

abouthiroppy:


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 2
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
```
icambron: Should be false.

icambron:
Looks good other than the one comment. Thanks.

abouthiroppy:
Updated;)

icambron:
Thanks!

PR #173, opened 2018-02-04
Fix math at the end of months

### LINKED ISSUE #167: Month math at the end of months
(assuming this is not a bug)
```
DateTime.local(2018, 3, 31).minus({months: 1}).toISO()
> "2018-03-03T00:00:00.000-03:00"
DateTime.local(2018, 1, 31).plus({months: 1}).toISO()
> "2018-03-03T00:00:00.000-03:00"
```

The only part of the documentation that can explain this behavior is:

> It's best not to think of calendar math as requiring arcane checks on the lengths of intervening periods. Instead, think of them as **adjusting that unit directly and keeping lower order date components constant.**

I think this could be more explicit, since it's different from other usual date manipulation implementations
[[1]](https://msdn.microsoft.com/library/system.datetime.addmonths(v=vs.110).aspx) [[2]](http://momentjs.com/docs/#special-considerations-for-months-and-years) [[3]](https://github.com/date-fns/date-fns/issues/17) [[4]](http://labix.org/python-dateutil#line-229)

rogerioyuuki:
Fixes #167

### DIFF
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -96,10 +96,13 @@ export class Util {
   }
 
   static daysInMonth(year, month) {
-    if (month === 2) {
-      return Util.isLeapYear(year) ? 29 : 28;
+    const modMonth = 1 + (((month - 1) % 12) + 12) % 12,
+          modYear = year + (month - modMonth) / 12;
+
+    if (modMonth === 2) {
+      return Util.isLeapYear(modYear) ? 29 : 28;
     } else {
-      return [31, null, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31][month - 1];
+      return [31, null, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31][modMonth - 1];
     }
   }
 
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -72,6 +72,15 @@ export class Util {
     return Util.isNumber(thing) && thing >= bottom && thing <= top;
   }
 
+  /**
+   * JS implementation (truncated division) of x % n takes the sign of the dividend (x).
+   * 
+   * This implementation takes the sign of the divisor (n)
+   */
+  static floorMod(x, n) {
+    return x - n * Math.floor(x / n);
+  }
+
   static padStart(input, n = 2) {
     return ('0'.repeat(n) + input).slice(-n);
   }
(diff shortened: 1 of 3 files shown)


test/datetime/math.test.js
```diff
@@ -18,11 +18,18 @@ function createDateTime() {
...
 // #plus()
 //------
-test('DateTime#plus({ year: 1}) adds a year', () => {
+test('DateTime#plus({ years: 1 }) adds a year', () => {
   const i = createDateTime().plus({ years: 1 });
   expect(i.year).toBe(2011);
 });
 
+test('DateTime#plus({ months: 1 }) at the end of the month', () => {
+  const i = DateTime.fromISO('2018-01-31T10:00'),
```
icambron: Can you also do a leap year?

icambron: Also, a test with `minus`

src/impl/util.js
```diff
@@ -96,10 +96,13 @@ export class Util {
   }
 
   static daysInMonth(year, month) {
-    if (month === 2) {
-      return Util.isLeapYear(year) ? 29 : 28;
+    const modMonth = 1 + (((month - 1) % 12) + 12) % 12,
```
icambron: This seems overly complicated. In what circumstance is that different from:

```js
1 + (month - 1) % 12;
```

I think the more complicated one is actually broken for 0:

```js
modMonth = (month) => 1 + (((month - 1) % 12) + 12) % 12
modMonth2 = (month) => 1 + (month - 1) % 12
modMonth(0) //=> 12, seems wrong
modMonth2(0) //=> 0, seems right
```

I *think* they both work out for negatives (e.g. `modMonth(-25) //=> 11`), but I want to see some tests for minus.

rogerioyuuki: Since in Luxon, months are 1-indexed, 0  is not a valid month. January is 1, so 0 should be December the previous year. (12)

Also, the simplified version doesn't work with negatives. 
 
`modMonth2(-25) //=> - 1`

test/datetime/math.test.js
rogerioyuuki: Sure, I'll write more tests!

src/impl/util.js
rogerioyuuki: I could refactor this expression into an `Util.positiveMod` or something like that, since I agree it looks more complicated than it needs to be. Some description of what this does would increase clarity.

icambron:
I'm going to take a look at this today, hopefully merge it, and it release it

src/impl/util.js
icambron: Ah, right, I got confused in the middle of that analysis. Cool.

PR #184, opened 2018-02-16
Fix incorrect format description in docs

codeclown:
Noticed this incorrect description while reading the docs.


icambron:
Thanks!

Issue #186, opened 2018-02-17
Interval#splitBy can cause infinite loop


vlad-zhukov:
`Interval#splitBy` hangs the process with an infinite loop if `duration` argument is not valid. 

Repro: https://codesandbox.io/s/n469pm30o4 (WARNING: this will hang the tab).

```js
const dateStart = DateTime.local().startOf("day");
const dateEnd = dateStart.plus({ days: 6 }).endOf("day");
const dates = Interval.fromDateTimes(dateStart, dateEnd).splitBy(DateTime.local());
```

icambron:
Yeah, we should fix that. Would make for a good first contribution!

vlad-zhukov:
I'll take it!

icambron:
@bogdanpolak Someone just needs to take #187 and finish the work. It needs passing tests.

bogdanpolak:
@ichernev I need little support, how to start this work. Should I clone current master and copy @vlad-zhukov changes into current sorce code? Then creating new pull request?

icambron:
Here's what I'd do:

1. Fork this repository and clone it
2. add vlad's fork as a remote and pull his branch
3. rebase it against master
4. push it to your fork
5. make a new PR

Does that make sense?

bogdanpolak:
@icambron Thx Issac for you assistance. I'm newbie in OpenSource on GitHub, but trying my best. 
For sure you suggestions make sense.
> 1. Fork this repository and clone it
forked: https://github.com/bogdanpolak/luxon and cloned.
> 2. add vlad's fork as a remote and pull his branch
Problem: can't find Vlad's fork.
Maybe this is an issue: vlad-zhukov deleted the vlad-zhukov:duration-toobject  branch 27 days ago?

icambron:
@bogdanpolak ah, right. Yes you'll have to make your own branch and integrate the changes manually then

bogdanpolak:
I just run luxon unit test in the cloned current main brunch and received 80 failed test. Should it looks that way? After small investigation I found that some failed tests are dependent on the OS X locale, I'm working on pl-pl locale settings. Maybe with docker will be better?

icambron:
Try `scripts/test`

bogdanpolak:
Now a little better: 35 failed. Those based on en-US locales was passed, but sill those which switching locales into different languages was failed (French, 
. Example:
>DateTime#weekdayLong returns the human readable weekday for fr locale
    expect(received).toBe(expected)
[...]

PR #187, opened 2018-02-17
Validate Duration.toObject() argument

vlad-zhukov:
Check that the argument is an object and has any units, otherwise create an invalid Duration.

Related to #186.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -6,7 +6,8 @@ import { Settings } from './settings';
 import { InvalidArgumentError, InvalidDurationError, InvalidUnitError } from './errors';
 
 const INVALID = 'Invalid Duration',
- UNPARSABLE = 'unparsable';
+  INVALID_INPUT = 'invalid input',
+  UNPARSABLE = 'unparsable';
 
 // unit conversion constants
 const lowOrderMatrix = {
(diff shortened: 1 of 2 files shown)


vlad-zhukov:
As I expected lots of tests failed but CI is green. The setup is clearly not working right.

icambron:
Arg, I've had this problem before: the tests are failing but the CI isn't reporting it. I'll see if I can sort that out.

icambron:
OK, if you rebase onto master, the CI automation should now report failure.

vlad-zhukov:
Rebased. So what do you think about checks like this? Should such methods return an invalid instance or simply throw an early error?

src/duration.js
```diff
@@ -197,8 +198,26 @@ export class Duration {
    * @return {Duration}
    */
   static fromObject(obj) {
+    if (obj == null || typeof obj !== 'object') {
+      return Duration.invalid(INVALID_INPUT);
```
icambron: This should throw. This is much more likely a programmer error than a data problem. It's similar to how we throw when the user specifies invalid units.

src/duration.js
```diff
@@ -197,8 +198,26 @@ export class Duration {
...
+
+    let hasUnits = false;
+    for (let i = 0, l = orderedUnits.length; i < l; i++) {
+      if (values[orderedUnits[i]]) {
+        hasUnits = true;
+        break;
+      }
+    }
+
+    if (hasUnits === false) {
```
icambron: flip the check:

```js
if (hasUnits) {
 return new Duration({
   values: 
   // ...
} else
  throw ...
}

File: src/duration.js [hunk 3]
```diff
@@ -197,8 +198,26 @@ export class Duration {
...
+      return Duration.invalid(INVALID_INPUT);
```
icambron: I think this is throwable too

src/duration.js
```diff
@@ -197,8 +198,26 @@ export class Duration {
...
   static fromObject(obj) {
+    if (obj == null || typeof obj !== 'object') {
+      return Duration.invalid(INVALID_INPUT);
+    }
+
+    const values = Util.normalizeObject(obj, Duration.normalizeUnit, true);
+
+    let hasUnits = false;
+    for (let i = 0, l = orderedUnits.length; i < l; i++) {
+      if (values[orderedUnits[i]]) {
```
icambron: I think we can just check if `values` has any `ownProperties`, right? `Object.keys(values).length === 0`

vlad-zhukov:
Thanks for comments, fixed. I am going to go through test errors soon and see what causes them.

icambron:
@vlad-zhukov just a reminder that this is still here

vlad-zhukov:
Hey! I'm not going to finish this because I don't use luxon anymore.

icambron:
okie-dokie

PR #192, opened 2018-02-19
Fix JSDocs

vlad-zhukov:


icambron:
The `@type` ones are right. See [https://esdoc.org/manual/tags.html#-type](here). Upper case object appears to be right too; see [here](https://esdoc.org/manual/tags.html#-return).

src/settings.js
```diff
@@ -15,7 +15,7 @@ let now = () => new Date().valueOf(),
 export class Settings {
   /**
    * Get the callback for returning the current timestamp.
-   * @type {function}
+   * @return {function}
```
icambron: I think you missed reverting these

src/duration.js
```diff
@@ -568,55 +568,55 @@ export class Duration {
 
   /**
    * Get the years.
-   * @return {number}
+   * @type {number}
```
icambron: 👍

vlad-zhukov:
Thanks you very much for helping! 🍕 I was loaded at work and forgot about this. 😞

PR #194, opened 2018-02-19
Move full-icu to dev deps

vlad-zhukov:
It's not used in sources.

This is critical for me because I can't install deps when I'm on a local network with a custom npm registry. `full-icu` runs a `post-install` script that tries to install additional deps, but it doesn't respect a custom registry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,6 @@
     "immutable"
   ],
   "repository": "https://github.com/moment/luxon",
-  "dependencies": {
-    "full-icu": "^1.2.0"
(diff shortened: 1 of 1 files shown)


icambron:
Weirdly, I have made this fix before. I'm not sure how that keeps creeping in there. It should definitely not be there.

vlad-zhukov:
Thanks! Would be great if you could publish a new minor update.

icambron:
yup, see 0.5.3

Issue #182, opened 2018-02-08
Viewing DateTime using React Dev Tools


fruitraccoon:
When a Luxon DateTime is passed as a prop to a React component, viewing the component props in the React Dev Tools shows the DateTime as expandable - however when expanded an `Uncaught TypeError: Cannot read property 'locale' of undefined` error appears in the console, indicating [this line](https://github.com/moment/luxon/blob/d485b16ba0aef87d2a9b9956f6c2593e0cfda16c/src/datetime.js#L779).

Digging into the Dev Tool code, this appears to be being caused by [the code](https://github.com/facebook/react-devtools/blob/40f926dfab63051446e9b671c369c703f2e21c67/agent/Bridge.js#L450) executing the `locale` property on the _prototype_ of DateTime, which has a `loc` of undefined.

Would a change to the `locale` prop code (and similar properties) to protect against an undefined `this.loc` be considered? Or perhaps there's a better way to solve the issue?

Thanks for this great library!

icambron:
Yeah, I'd take a PR checking if `loc` is undefined. I hadn't bothered because I didn't know of a scenario where it would matter.

icambron:
Fixed in #206

PR #206, opened 2018-03-10
Ensure prototype getters can be addressed without throwing

fruitraccoon:
Updates that allow Luxon classes to be expanded in the React dev tools. This addresses [#182](https://github.com/moment/luxon/issues/182)

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -779,7 +779,7 @@ export class DateTime {
    * @type {string}
    */
   get locale() {
-    return this.loc.locale;
+    return this.isValid ? this.loc.locale : null;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/datetime.js
```diff
@@ -797,15 +797,15 @@ export class DateTime {
...
-    return this.loc.outputCalendar;
+    return this.isValid ? this.loc.outputCalendar : null;
   }
 
   /**
    * Get the name of the time zone.
    * @type {string}
    */
   get zoneName() {
-    return this.invalid ? null : this.zone.name;
```
fruitraccoon: This getter also needed to be updated, as in the prototype `invalid` is `undefined`, thus falsy. But as `isValid` explicity checks for `null`, using it works just fine. Using `isValid` also matches the other getters that did not need to be updated.

icambron:
Thanks!

PR #223, opened 2018-03-27
fix npm script reference table for correct rendering

32graham:
Github and many other markdown processors won't render the npm script reference table correctly as it was. This makes it render correctly in Github and others.


icambron:
thanks!

Issue #226, opened 2018-04-05
Expose zones vs internal caching of zones


mhverbakel:
Hello,

After figuring out a way to speed up the library (https://github.com/moment/luxon/issues/89#issuecomment-378826414) I noticed that the `IANAZone` is not exposed by the luxon.js file. Can this be added, so I can use this in my code?

Thanks in advance!

mhverbakel:
Same goes for the other zone types (`LocalZone.instance`, `FixedOffsetZone`, `FixedOffsetZone.utcInstance`, maybe even `InvalidZone`?)

mhverbakel:
After https://github.com/moment/luxon/issues/89#issuecomment-378985031, the discussion on whether zones should be exposed or just internally cached can be held.

# Expose zones
There are two flavours of exposing the zones, either by having static members on the Zone class, or by exposing the real types in the library.
## PROS
- User can choose how and whether he wants zones cached
- Makes sense with the documentation (since `zone` accepts `string | Zone`, but you can never instantiate a `Zone` outside the library)
## CONS
- User must know that this can be used to improve performance, otherwise he won't benefit from it

# Internal caching
## PROS
- Works by default
## CONS
- User has no say in the method of caching. If he uses a lot of timezones, this might become a memory leak. Or are we gonna store only the last timezone? Or the last `x` timezones?

My preference goes toward having the zones exposed, but I'm willing to make a pull request for either of those options.

icambron:
Internal caching for sure. Luxon should be fast by default. Let's find out how much memory it takes to cache literally all of them. If it's a lot, then we'll need an LRU. If it's not, we shouldn't worry about it.

icambron:
See the branch `cache-iana-zones`. On my machine:

Before:
```
DateTime#setZone x 2,867 ops/sec ±3.18% (87 runs sampled)
```

After:
```
DateTime#setZone x 38,230 ops/sec ±4.36% (74 runs sampled)
```

So about 14x faster. Still not the fastest thing because it still does some Intl work to compute the offset. 

What I haven't done is check how much memory that eats up.

Can you give it a spin?

FWIW, I'm not really against exposing the zones. I just think Luxon should be fast without extra work by the user.

PR #227, opened 2018-04-05
Added exposes for specific zones.

mhverbakel:
Fixes https://github.com/moment/luxon/issues/226


icambron:
Thanks, I've decided this makes sense to do, even though I don't want it to be the main avenue for perf

PR #230, opened 2018-04-16
Add weeksInWeekYear Method

### LINKED ISSUE #205: weeksInYear
Hi.

Would be handy to include a weeksInYear property

DateTime.local(2018, 1, 1).weeksInYear //=> 52 
DateTime.local(2019, 1, 1).weeksInYear //=> 52 
DateTime.local(2020, 1, 1).weeksInYear //=> 53 

Apologies if this already exists.

Many thanks,

Luke.

kivlor:
Fixes #205

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -15,6 +15,7 @@ import {
   daysInMonth,
   daysInYear,
   isLeapYear,
+  weeksInWeekYear,
   normalizeObject
 } from './impl/util';
 import { normalizeZone } from './impl/zoneUtil';
@@ -1071,6 +1072,17 @@ export default class DateTime {
    */
   get daysInYear() {
     return this.isValid ? daysInYear(this.year) : NaN;
+  }
+
+  /**
+   * Returns the number of weeks in this DateTime's year
+   * @see https://en.wikipedia.org/wiki/ISO_week_date
+   * @example DateTime.local(2004).weeksInWeekYear //=> 53
+   * @example DateTime.local(2013).weeksInWeekYear //=> 52
+   * @type {number}
+   */
+  get weeksInWeekYear() {
+    return this.isValid ? weeksInWeekYear(this.year) : NaN;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


icambron:
Thanks for the PR. Some notes:

 * You only need one of the three bulleted definitions in the wikipedia article, not all three, if I understand the phrasing correctly ("can be described by any of the following equivalent definitions")
 * The rules themselves are super confusing. Do they mean the matching Gregorian year? That's how you interpreted it and that may well be right, but that's now how I read it. Jan 1 XXXX is likely the not the first day of the XXXX week year; it might not even be in the XXXX week year at all.
 * I'd rather do this without creating a new date instance (FWIW, also, you just wanted `new Date([year])`; you don't have to parse).

However, I think I can make this way easier! I thought that section of Wikipedia just below the rules you implemented looked familiar. Take a look at this: https://github.com/moment/luxon/blob/e77f74f27fae1237a5b59b1df15e4114215f665c/src/impl/conversions.js#L11. That's a direct implementation of those rules. Right now, `conversions` is just using that locally to do some conversion math, so it's not in the right place to be useful to us, but changing that looks to be straightforward:

 1. Move that function to `util`, renaming it `weeksInWeekYear`
 2. Import it to `conversions` so it can keep using it
 3. Call the new util method in `DateTime#weeksInWeekYear`
 4. Profit

Sound good?

kivlor:
Hey @icambron, thanks for the feedback! You're probably right in your assumptions on the Wiki article so I've gone ahead and implemented your simpler approach 😄 

I did try to get the second example working but failed... Turns out I was missing the Math.floor around each of the divisions 😓

icambron:
This looks great! Thanks

icambron:
Oops, I merged too fast! There's still one problem. Since it's not released, it's probably easier to fix with a new PR than to revert. I'll mark the problem in the diff.

src/datetime.js
```diff
@@ -1073,6 +1074,17 @@ export default class DateTime {
...
 
+  /**
+   * Returns the number of weeks in this DateTime's year
+   * @see https://en.wikipedia.org/wiki/ISO_week_date
+   * @example DateTime.local(2004).weeksInWeekYear //=> 53
+   * @example DateTime.local(2013).weeksInWeekYear //=> 52
+   * @type {number}
+   */
+  get weeksInWeekYear() {
+    return this.isValid ? weeksInWeekYear(this.year) : NaN;
```
icambron: `this.year` isn't right, and will break for dates with different ISO and Gregorian years. You need `this.weekYear`.

kivlor:
@icambron gahhh, no worries. I've replaced year with weekYear in #231

PR #232, opened 2018-04-17
Fixed month unit conversion bug

Cultrarius:
Looks like a copy-paste error from the year unit conversion.


icambron:
Wild. Apparently no has ever used this features and it has no tests :(.

icambron:
Thanks. I'll add some tests for this...eventually.

icambron:
Released in 1.1.0

PR #239, opened 2018-04-24
Update lockfile to match package.json versions

razor-x:
I am going to debug an issue my team may have found related to quarter math, but after running `npm install` the pacakge-lock.json file was updated. This indicates a mismatch between the versions in package.json and the lockfile. This is the result after running `npm install` on npm version 5.6.0 and Node 9.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -402,12 +402,6 @@
         "lodash": "4.17.4"
       }
     },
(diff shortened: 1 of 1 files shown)


razor-x:
I've updated this to use the lockfile generated by the included Docker script: `./docker/npm install` instead of my local version.

Issue #267, opened 2018-06-04
fromMillis question


benmccann:
https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#static-method-fromMillis says:

> Create an DateTime from a count of epoch milliseconds. Uses the default zone.

What does it mean to use the default zone? Is milliseconds since the epoch in the default zone? Or is it milliseconds since the epoch in UTC converted to the default zone?

I'm hoping it's the latter for consistency and ease of use. The value returned by JavaScript's `getTime` method is `the number of milliseconds since 1 January 1970 00:00:00 UTC`. Java's `System.currentTimeMillis()` returns `the difference, measured in milliseconds, between the current time and midnight, January 1, 1970 UTC`. I'd like to be able to take these values and create a `DateTime`.

icambron:
Yup, it's the second one. The millis always mean the same thing (ms since epoch UTC); it's just a question of what zone we set as the DateTime's zone. The docs could definitely be clearer there.

PR #268, opened 2018-06-05
Clarify fromMillis doc

benmccann:
I was confused by "default zone" whether it was the epoch in the default zone or the epoch in utc converted to the default zone. This clarifies it

Thank you for answering my question in https://github.com/moment/luxon/issues/267!


icambron:
Looks good, thanks

PR #275, opened 2018-06-17
Make DateTime#valueOf call #toMillis

shvaikalesh:
Previous behavior: `toMillis` calls `valueOf`.

This PR aligns Luxon's `DateTime#valueOf` with it's very own `Duration#valueOf` (introduced in https://github.com/moment/luxon/pull/261), [implementation in moment.js](https://github.com/moment/moment/blob/master/src/lib/moment/to-type.js#L1-L3), and the native [`Date#valueOf`](https://tc39.github.io/ecma262/#sec-date.prototype.valueof). They all have logicless `valueOf` that just calls some other method. And no other method calls `valueOf` directly, only JavaScript does.


test/datetime/invalid.test.js
```diff
@@ -39,7 +39,7 @@ test('Invalid tell you why', () => {
 });
 
 test('Invalid DateTimes return invalid Dates', () => {
-  expect(organic1.toJSDate().valueOf()).toBeFalsy();
+  expect(organic1.toJSDate().valueOf()).toBe(NaN);
```
icambron: Surprised this works, actually.

icambron:
This looks good. The build error is probably because I haven't pinned down the version of prettier. I bet if you `npm update`, then `npm run format`, then push that up, it will resolve. I'll fix that issue soon.

test/datetime/invalid.test.js
shvaikalesh: It didn't work initially, was fixed by https://github.com/facebook/jest/pull/4917, that's why I updated `jest`.

PR #280, opened 2018-06-21
Clarify a comment

shvaikalesh:
While I empathize the original comment,
1) I firmly believe that Luxon will influence an ongoing [Temporal Proposal](https://github.com/tc39/proposal-temporal) effort, thus we should keep the code base clean of harsh comments. TC39 people are doing an amazing work.
2) JavaScript had to include date API and within very tight deadline. It also was fundamental to be Java-like, to boost the adoption. `java.util.Date` had issues, but it was a very reasonable call. Without any of that, we probably wouldn't have a JavaScript community as we know it.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -123,7 +123,7 @@ function objToLocalTS(obj) {
     obj.millisecond
   );
 
-  // javascript is stupid and i hate it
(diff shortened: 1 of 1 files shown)


icambron:
lol, sure

icambron:
thanks!

PR #281, opened 2018-06-22
Use Symbol for Node's custom inspect

### LINKED ISSUE #276: DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated on Node v10+
```
//index.mjs
import luxon from 'luxon'; // import { DateTime } from 'luxon' not working yet (https://github.com/moment/luxon/issues/252)
const DateTime = luxon.DateTime;
console.log(DateTime.local())
```

=>

```
$ node -v
v10.4.1
$ node --trace-deprecation --harmony --experimental-modules index.mjs
DateTime {
  ts: 2018-06-17T11:28:08.858+02:00,
  zone: Europe/Rome,
  locale: en-US }
(node:99259) [DEP0079] DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated
    at formatValue (util.js:467:31)
    at inspect (util.js:328:10)
    at Proxy.formatWithOptions (util.js:182:12)
    at Console.(anonymous function) (console.js:188:15)
    at Console.log (console.js:199:31)
    at Object.log (/Users/damz/Desktop/yarsk/node_modules/better-console/index.js:22:17)
    at file:///Users/damz/Desktop/yarsk/index.mjs:30:9
    at ModuleJob.run (internal/modules/esm/module_job.js:106:14)
```

shvaikalesh:
Resolves #276.

Node's `inspect` is fairly useful, let's try to keep it. I don't exceptionally like the solution at hand, but I have an idea how we can make it better if Node.js would support it.

### DIFF
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -203,3 +203,11 @@ export function normalizeObject(obj, normalizer, ignoreUnknown = false) {
 export function timeObject(obj) {
   return pick(obj, ['hour', 'minute', 'second', 'millisecond']);
 }
+
+export const customInspectSymbol = (() => {
+  try {
+    return require('util').inspect.custom; // eslint-disable-line global-require
+  } catch (_err) {
+    return Symbol('util.inspect.custom');
+  }
+})();
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -16,6 +16,7 @@ import {
   daysInYear,
   isLeapYear,
   weeksInWeekYear,
+  customInspectSymbol,
   normalizeObject
 } from './impl/util';
 import { normalizeZone } from './impl/zoneUtil';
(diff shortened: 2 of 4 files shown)


src/impl/util.js
```diff
@@ -203,3 +203,11 @@ export function normalizeObject(obj, normalizer, ignoreUnknown = false) {
 export function timeObject(obj) {
   return pick(obj, ['hour', 'minute', 'second', 'millisecond']);
 }
+
+export const customInspectSymbol = (() => {
+  try {
+    return require('util').inspect.custom; // eslint-disable-line global-require
+  } catch (_err) {
+    return Symbol('util.inspect.custom');
```
icambron: Can you explain this? I'm a little out of the loop on this stuff

shvaikalesh: Oh, sure.

The recommended way for custom routine during `inspect` is to specify a symbol, which is exposed as `util.inspect.custom`.

However, for browser environments, where `require` is unavailable, we need a clean no-op. That is handled in `catch` clause: `util.inspect.custom` is just for show, it is merely a description, could be anything. We simply return a new symbol (that won't be used by anyone) so we can simply do

```js
import { customInspectSymbol } from 'utils';

[customInspectSymbol]() {
  // ...
}
```

in our classes, without any helpers or extra checks.

icambron: Got it

icambron:
thanks!

PR #285, opened 2018-06-28
Replace all `instanceof Object` checks

rricard:
Those assertions fail (even if we're giving actual objects) while working across different JS contexts.

Not sure that `typeof` is the best option for that though. I would just remove that check altogether since there are other subsequent checks in `fromObject` that will fail if the thing is not an object.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1860,7 +1860,7 @@ export function friendlyDateTime(dateTimeish) {
     return dateTimeish;
   } else if (dateTimeish.valueOf && isNumber(dateTimeish.valueOf())) {
     return DateTime.fromJSDate(dateTimeish);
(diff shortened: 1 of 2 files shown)


icambron:
> there are other subsequent checks in fromObject that will fail if the thing is not an object

Do they fail in helpful ways, though?

rricard:
For durations, yes, we have the same exact check: https://github.com/moment/luxon/blob/b8b8a9b1da0637d101db6b8cae43f33de4fefeaa/src/duration.js#L241

For the DateTime, not that I know of

rricard:
@icambron is there anything I need to change here to get that PR potentially merged? Thanks!

icambron:
Thanks

PR #301, opened 2018-07-18
Implemented fromMillis() to verify args

### LINKED ISSUE #300: Check the milliseconds argument of fromMillis
I was not paying attention when I passed a `DateTime` object to the `fromMillis` method.
The resulting date seemed to be valid, except it doesn't: its internal value gets populated with the `DateTime` instance, breaking any subsequent comparison or calculation.

I would expect `fromMillis` to return an invalid date.

Some tests from the curent version (`v1.3.1`):

```javascript
const DateTime = require('luxon').DateTime;
const now = DateTime.local();
const lastYear = now.minus({ years: 1 });
const wrongDate = DateTime.fromMillis('now');
const testDate = DateTime.fromMillis(now);

/* Expected*/
console.log(now.isValid && now.toISO()); // 2018-07-17T14:54:22.448+02:00
console.log(lastYear.isValid && lastYear.toISO()); // 2017-07-17T14:54:22.448+02:00
console.log(now.diff(lastYear).as('days')); // 365
console.log(now > lastYear); // true

/* Not expected */
console.log(testDate.isValid && testDate.toISO()); // 2018-07-17T14:54:22.448+02:00
console.log(wrongDate.isValid && wrongDate.toISO()); // 0NaN-NaN-NaNTNaN:NaN:NaN.NaN-NaN:NaN
console.log(now.equals(testDate)); // false
console.log(testDate > lastYear); // false
console.log(testDate.diff(lastYear).as('days')); // 0
console.log(testDate.valueOf() instanceof DateTime && testDate.valueOf().toISO()); // 2018-07-17T14:54:22.448+02:00
```

taeukme:
Fixes #300

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -501,11 +501,15 @@ export default class DateTime {
    * @return {DateTime}
    */
   static fromMillis(milliseconds, options = {}) {
-    return new DateTime({
-      ts: milliseconds,
-      zone: normalizeZone(options.zone, Settings.defaultZone),
-      loc: Locale.fromObject(options)
-    });
+    if (!isNumber(milliseconds)) {
+      throw new InvalidArgumentError('fromMillis requires a numerical input');
+    } else {
+      return new DateTime({
+        ts: milliseconds,
+        zone: normalizeZone(options.zone, Settings.defaultZone),
+        loc: Locale.fromObject(options)
+      });
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


icambron:
Looks good, but needs a test

taeukme:
The tests are up but coverage seems to be failing.

icambron:
Yeah, the coverage is failing because the tests never follow the fail code path. We already have tests for `fromMillis`, but you need one for the fail case, like `expect(DateTime.fromMillis('goat').toThrow(...`

icambron:
Awesome. Thanks for the contribution!

PR #319, opened 2018-08-16
Add X and x tokens to toFormat that return the Date in epoch seconds and milliseconds respectively

ivosabev:
Let me know if this is of interest and I will add the needed docs to the pull request.


icambron:
First, sorry for the ridiculously long delay on getting to this.

A few thoughts:

I don't like `toUnix()` since it doesn't seem to add much above `toMillis()/1000`. The new token is a different story; I can see how it's painful to handle that in userland. I'm thus going to close your other PR with both `toUnix` and the new token in favor of this one.

To get this over the line, we just need docs and tests.

ivosabev:
Added the two tokens `x` and `X` + tests ands docs. Let me know if there is anything else needed.

icambron:
Looks great. Thanks for the contribution.

I'll release this soon.

dasa:
Could these be supported in `fromFormat` as well?

Issue #302, opened 2018-07-22
Etc/GMT timezones are parsed incorrectly


palamccc:
Timezones like Etc/GMT+8 should be converted to UTC-8 But they are parsed as UTC+8.
I verified this against moment timezone database, and also with another SQL database. Moment already handled this  https://github.com/moment/moment-timezone/issues/191

```js
const moment = require('moment-timezone');
const { DateTime } = require('luxon');

console.log(DateTime.fromISO('2018-07-15', { zone: 'Etc/GMT+8' }).setZone('UTC').toISO());
console.log(moment.tz('2018-07-15', 'Etc/GMT+8').tz('UTC').toISOString());
```
Output
```
2018-07-14T16:00:00.000Z
2018-07-15T08:00:00.000Z
```
- runkit https://runkit.com/palamccc/5b0f89810acbb10012e61d20

nvtin:
Hi @icambron Any fixing for it?
I am using luxon in our project and I see that there are some thing was not correct with timezone.

I just go deeply to your test file [zone.test.js line 202](https://github.com/moment/luxon/blob/master/test/datetime/zone.test.js#L202)
It look there are some confused there. The Etc/GMT+8 should be converted to UTC-8 But they are parsed as UTC+8 as in your test. 

Can you have a look at it?
Hope it will be fixed soon. Thanks @icambron

icambron:
@nvtin I would certainly accept a pull request

icambron:
Fixed in #327, will release soon

PR #327, opened 2018-09-21
bug fix for Etc/GMT timezones

palamccc:
Fix for https://github.com/moment/luxon/issues/302, also I couldn't run tests on my local machine. so didn't check whether all tests are passing.


icambron:
Not sure what's up with the build, but once I have that fixed, I'll merge this. Thanks for the contribution!

PR #341, opened 2018-10-06
Interval.fromISO to allow relative durations.

### LINKED ISSUE #340: Interval.fromISO does not allow a duration relative to a start or end.
[ISO 8601 allows for intervals to be defined](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals) as 
 - `<start>/<end>`
 - `<start>/<duration>`
 - `<duration>/<end>`

But only `<start>/<end>` is currently supported.

```js
const i = luxon.Interval.fromISO("2007-03-01T13:00:00Z/P1Y2M10DT2H30M")
i.isValid // false
```

If there is interest in adding this support I'd be happy to open a PR updating this block to check if both `DateTime`s are valid and calling `Duartion.fromISO` then combining.
https://github.com/moment/luxon/blob/747d32719042d7ff05716e83372a902b41e229b6/src/interval.js#L109-L111

It's unclear if this goes against [# 3 of the design philosophy](https://github.com/moment/luxon/blob/master/contributing.md), my code could do the work of splitting on `/` and call `DateTime`, `Duration`, and `Interval.fromDateTimes` manually. But my understanding of ISO 8601 is that a value like `2007-03-01T13:00:00Z/P1Y2M10DT2H30M` is a single interval value and in that case having this library do the work seems reasonable.

mastermatt:
Fixes #340.

Start accepting `<start>/<duration>` and `<duration>/<end>` formats of [ISO 8601 intervals](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals).

One very minor breaking change is that an invalid input with a slash now
has a `invalidReason` of `invalid ISO format` instead of `invalid endpoints`.

```js
const i = Interval.fromISO("foo/bar");
i.invalidReason; // invalid ISO format
```

### DIFF
--- a/src/interval.js
+++ b/src/interval.js
@@ -98,16 +98,33 @@ export default class Interval {
   }
 
   /**
-   * Create an Interval from an ISO 8601 string
+   * Create an Interval from an ISO 8601 string.
+   * Accepts `<start>/<end>`, `<start>/<duration>`, and `<duration>/<end>` formats.
    * @param {string} string - the ISO string to parse
-   * @param {Object} opts - options to pass {@see DateTime.fromISO}
+   * @param {Object} [opts] - options to pass {@link DateTime.fromISO} and optionally {@link Duration.fromISO}
+   * @see https://en.wikipedia.org/wiki/ISO_8601#Time_intervals
    * @return {Interval}
    */
   static fromISO(string, opts) {
-    if (string) {
-      const [s, e] = string.split(/\//);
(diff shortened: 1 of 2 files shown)


mastermatt:
@icambron I'm not sure what's up with Travis. The new tests and linter runs fine locally.

icambron:
This looks good to me. I'll get it merged when I figure out what's screwed up with the build system.

PR #344, opened 2018-10-08
Fixes DateTime.toRFC2822 using 12h instead of 24h format

juancampa:
According to [RFC2822](https://www.ietf.org/rfc/rfc2822.txt) hours must be in 00-23 range. Current implementation was using `hh` which is 00-12. 

This was causing, for example a 3pm and a 3am date to print the same RFC2822 value:

```
> DateTime.local().set({ hour: 3 }).toRFC2822()
'Mon, 08 Oct 2018 03:53:26 -0400'
> DateTime.local().set({ hour: 15 }).toRFC2822()
'Mon, 08 Oct 2018 03:53:26 -0400'
```

With this change, the second one is correctly printed to have hour 15.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1432,7 +1432,7 @@ export default class DateTime {
    * @return {string}
    */
   toRFC2822() {
-    return toTechFormat(this, "EEE, dd LLL yyyy hh:mm:ss ZZZ");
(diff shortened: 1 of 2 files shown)


juancampa:
Tests are passing, linting is failing for some reason. I had the same problem locally. I don't think is related to my changes though.

icambron:
This looks good. Thanks!

PR #358, opened 2018-11-03
Testing out alternative type checks for Duration and DateTime.

icambron:


icambron:
Hmm, need to open a new PR to pull in my changes.

PR #371, opened 2018-11-20
Add support for Unix timestamps in seconds.

peterkhayes:
This pull request adds `toSeconds` and `fromSeconds` methods for working with Unix timestamps. 

Now, I saw that https://github.com/moment/luxon/pull/319 added support for timestamps inside the `.toFormat` function, but didn't add explicit timestamp support. And, I've read the contribution guidelines, which state that:

> Luxon shouldn't contain simple conveniences that bloat the library to save callers a couple lines of code. Write those lines in your own code.

However, I think supporting Unix timestamps is an important use case. This is because many other systems and specifications use Unix timestamps in seconds as their primary date format. For example:

- Python's `time()`
- Ruby's `Time.now.to_i`
- Bash's `date +%s`
- C's `time(null)`

Because of this, many APIs specify timestamps in seconds. Currently, working with these times is a bit cumbersome, and prone to errors.

```js
const inputSeconds = 1542674993;
const inputDT = DateTime.fromMillis(inputSeconds * 1000);

const outputDT = dt.plus({days: 1});
const outputSeconds = Math.floor(dt.toMillis() / 1000);
// or
const outputSeconds = Number(dt.toFormat("X"));
```

It's easy to forget to multiply or divide by 1000, or to cast the `toFormat` result as a string, and this can lead to subtle bugs. I think having built-in support for seconds is pretty light-weight and makes working with timestamps much easier. An example use case is comparing a Luxon DateTime object to an array of GeoJSON features, each of which has numeric `timestamp` property.

We're currently migrating from Moment to Luxon, which has been really nice, except for this. Moment supports unix timestamps directly with `.unix()`, and adding multiplications/divisions everywhere where we used `.unix()` feels less than elegant.

Anyway, I understand your concerns about bloating the library, but it'd be really nice to see this get in.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -520,6 +520,28 @@ export default class DateTime {
   }
 
   /**
+   * Create a DateTime from a number of seconds since the epoch (i.e. since 1 January 1970 00:00:00 UTC). Uses the default zone.
+   * @param {number} seconds - a number of seconds since 1970 UTC
+   * @param {Object} options - configuration options for the DateTime
+   * @param {string|Zone} [options.zone='local'] - the zone to place the DateTime into
+   * @param {string} [options.locale] - a locale to set on the resulting DateTime instance
+   * @param {string} options.outputCalendar - the output calendar to set on the resulting DateTime instance
+   * @param {string} options.numberingSystem - the numbering system to set on the resulting DateTime instance
+   * @return {DateTime}
+   */
+  static fromSeconds(seconds, options = {}) {
+    if (!isNumber(seconds)) {
+      throw new InvalidArgumentError("fromSeconds requires a numerical input");
+    } else {
+      return new DateTime({
+        ts: seconds * 1000,
+        zone: normalizeZone(options.zone, Settings.defaultZone),
+        loc: Locale.fromObject(options)
+      });
+    }
+  }
+
+  /**
    * Create a DateTime from a Javascript object with keys like 'year' and 'hour' with reasonable defaults.
    * @param {Object} obj - the object to create the DateTime from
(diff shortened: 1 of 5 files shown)


icambron:
I had been hoping to avoid adding this one, but it's been asked for a bunch of times now, so I'm going to say yes.

src/datetime.js
```diff
@@ -1552,6 +1574,14 @@ export default class DateTime {
     return this.isValid ? this.ts : NaN;
   }
 
+  /**
+   * Returns the epoch seconds of this DateTime.
+   * @return {number}
+   */
+  toSeconds() {
+    return this.isValid ? Math.floor(this.ts / 1000) : NaN;
```
icambron: Should we really be flooring here? Why not just return it with the decimals? (I think this really is a different case than the `toFormat` one)

peterkhayes: Yeah, I went back and forth on this one. If you think this is better as a float, I'm happy to change it.

peterkhayes: I amended my commit to not round seconds.

icambron:
Cool. This is good to go. Thanks for the contribution!

icambron:
I'll release this soon, probably as 1.8.0

jaripekkala:
The 1.8.0 release change log says it includes these new methods, however it seems like it doesn't. Very confusing, am I missing something?

icambron:
@jaripekkala Yeah, you're right. Weird. I'll figure out what happened there.

icambron:
@jaripekkala Not sure what happened, but my releasing tooling must have done something funky. Anyway, try 1.8.1

PR #379, opened 2018-12-08
Fixed output examples for toLocaleString

henningta:
Tested using en-US locale, which I think is what the example outputs are using.


icambron:
thanks!

PR #383, opened 2018-12-13
Fix Issue 376

vin-car:
Fixed `isDateTime` and `isDuration` to make them return `false` instead of `undefined`.


icambron:
Looks great, thanks! I'll release this soon.

PR #387, opened 2018-12-21
Added aliases for zoneName and setZone

webdevium:
- method `setTz()` is alias for method `setZone()`
- property `tz` is alias for property `zoneName`
- added some test cases for new functionality
- additionally, imported **package-lock.json** to **yarn.lock**


icambron:
Thanks for offering this, but I don't think I want these aliases. `setTz` only saves two characters. `tz` is ambiguous about what it returns. The existing names seem fine; what motivates your need for these aliases?

icambron:
Sorry, not convinced.

PR #390, opened 2018-12-26
Relative time formatting

icambron:


icambron:
This still needs some work. Specifically, the way it handles `DateTime.local().plus({ days: 1 }).fromNow()` will return "in 24 hours" or "tomorrow" based on whether or not a millisecond passed before it managed to create the "now" time. Trying to decide whether we should add some thresholding or just make that "in 23 hours".

PR #396, opened 2019-01-04
Support for midnight or 24:00

ukstv:
ISO8601 treats midnight in a specific way. It treats 00:00 as start of the day. 24:00 is 00:00 of the next day, or the end of the current day. Luxon prohibits using 24:00 totally.

Case to check is DateTime.fromISO('24:00'). For today (2018-01-04) it should return "2018-01-05 00:00".

This is of utter importance for intervals. One should be able to specify an interval till the end of the day. Interval.fromIso('23:00/24:00') currently throws an error, while it is expected to be an hour before midnight.

### DIFF
--- a/src/impl/conversions.js
+++ b/src/impl/conversions.js
@@ -137,7 +137,7 @@ export function hasInvalidGregorianData(obj) {
 }
 
 export function hasInvalidTimeData(obj) {
-  const validHour = numberBetween(obj.hour, 0, 23),
+  const validHour = numberBetween(obj.hour, 0, 24),
     validMinute = numberBetween(obj.minute, 0, 59),
     validSecond = numberBetween(obj.second, 0, 59),
     validMillisecond = numberBetween(obj.millisecond, 0, 999);
--- a/test/datetime/regexParse.test.js
+++ b/test/datetime/regexParse.test.js
@@ -475,6 +475,18 @@ test("DateTime.fromISO() accepts hour:minute", () => {
   });
 });
 
+test("DateTime.fromISO() accepts 24:00", () => {
+  isSame("2018-01-04T24:00", {
+    year: 2018,
+    month: 1,
+    day: 5,
+    hour: 0,
+    minute: 0,
+    second: 0,
+    millisecond: 0
+  });
+});
+
 test("DateTime.fromISO() accepts some technically incorrect stuff", () => {
   // these are formats that aren't technically valid but we parse anyway.
   // Testing them more to document them than anything else


src/impl/conversions.js
```diff
@@ -137,7 +137,7 @@ export function hasInvalidGregorianData(obj) {
 }
 
 export function hasInvalidTimeData(obj) {
-  const validHour = numberBetween(obj.hour, 0, 23),
```
icambron: Hmm, does this work in general? This permits lots of other methods to be passed 24 hours, which might be OK but needs thought through

ukstv: It does indeed. After some thought, it is totally ok (and in compliance with the standard) for the hour to be 24 as long as it is parsed correctly into `00:00` for the next day. I could not find a case where parsing fails, so I believe it is all right.

I am still confused by the fact this simple change fixes the handling of midnight everywhere. Apparently it does for `DateTime`, and `Interval` both.

icambron: Yeah, it works because the native Date constructor handles overflows in the hour position.

It has a lot of other implications:

```js
// this is a spec violation, as I understand it. 24:00 is ok but I don't think 24:30 is
// I think that's probably ok to let slide, since it's better to get more valid strings included than
// be strict about the policing bad strings
> DateTime.fromISO("2019-01-09T24:30:30").toISO()
'2019-01-10T00:30:30.000-05:00'

// this is fine and probably a good thing:
> DateTime.fromObject({ year: 1999, month: 12, day: 31, hour: 24 }).toISO()
'2000-01-01T00:00:00.000-05:00'

// This is a little wild: Nov 6, 1994 is a Sunday, which means this will complain that the day is wrong:

> DateTime.fromHTTP("Sun, 06 Nov 1994 24:00:00 GMT").invalidReason
'mismatched weekday'

//whereas Nov 6 is not Monday, but this passes because it's really parsing Monday:
> DateTime.fromHTTP("Mon, 06 Nov 1994 24:00:00 GMT").isValid
true

// (also, 24 is a spec violation for HTTP dates, which also seems minor, but still, starting to creep here)
```

Those were the only things I could find just playing with it. I think most of those are probably fine, though I may make a few changes to lock it down a bit more (it wouldn't be very hard to make it require that if the hour is 24, that the minutes, seconds, and milliseconds be zero, or to isolate the change to ISO parsing)

icambron:
Sorry for dragging my feet on this. I have an outstanding todo item here. I will get to it asap

icambron:
going to merge these changes in #444

PR #397, opened 2019-01-05
Update datetime.js

tonysepia:
I have been using prototype extentions in my code to make luxon objects compatible with various components that interrogate standard JS date objects (Angular's formatDate filter, uib-datepicker, etc)
Would the community be interested in adding them to the component by default to increase compatibility with JS component ecosystem, without resorting to toJSDate, which strips a lot of information from the object?

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1812,20 +1812,20 @@ export default class DateTime {
       })
     );
   }
-  
+
   /**
-   * Returns the time zone difference, in minutes, from current locale (host system settings) to UTC. 
-   * Mimics behaviour of getTimezoneOffset in JavaScript, improving compatibility with frameworks 
+   * Returns the time zone difference, in minutes, from current locale (host system settings) to UTC.
+   * Mimics behaviour of getTimezoneOffset in JavaScript, improving compatibility with frameworks
    * and UI components that rely on this method (i.e. Angular dateFormat filter, date pickers, etc.)
    */
   getTimezoneOffset() {
     return DateTime.local().offset * -1;
   }
-  
- /**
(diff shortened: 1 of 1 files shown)


icambron:
Can you explain this more? These are needed for places where you pass the Luxon DateTime into a method that expects a native Date? If so, why these particular two methods and not the rest of Date's interface?

tonysepia:
Hi Isaac,

What do you mean when you say "why are these methods not in Dates
interface"? They are part of Date, but not luxon Datetime.

When we get an ISO date string from a model, we convert it into luxon
object (because of all the great advantages luxon has over JS Date).

This luxon object is now part of our data binding: various elements of the
interface read its value and format it for displaying to the user. These
parts of the interface are designed to work with JS Date and interrogate
some if it's methods that luxon doesn't have. And it doesn't make sense to
create another copy of this date variable just to feed the interface
elements.

I hope this makes sense!

Kind regards,

icambron:
Hmm, I think you misunderstood my question. I said "If so, why these particular two methods and not the rest of Date's interface?" What I mean is: there are lots of method on Date that aren't in Luxon's DateTime. I wouldn't want to commit to having Luxon mirror the whole Date interface, and I'm struggling to see why the particular two you added are special.

icambron:
I don't think Luxon will take this contribution. Here are my thoughts:

 * To be useful to people, I'd have to establish that Luxon is supporting the Date interface, which is unrealistic.
 * I understand why adding those extension methods is useful, and those prototype extensions is indeed the right way to do it. If that's broadly useful, it would make a great external library.
 * In general, I think people who want Date objects should just call `toJSDate()`

Issue #386, opened 2018-12-20
Interval.setZone() / Interval.toUTC()


pietrovismara:
I find myself having often to change an `Interval` `DateTime`s zone.

This usually involves me doing something like:

```
interval = interval.set({
    start: int.start.setZone(zone),
    end: int.end.setZone(zone)
});

```

As a small usability improvement, it would be nice to have the `setZone` and `toUTC` methods on `Interval`s, which would allow more concise code like the following:

`interval = interval.setZone(zone);`

What is your opinion about it?

icambron:
It's a little to specific, and I worry it's a slippery slope to adding the whole DateTime interface to Interval.

What if instead we added `bothEndpoints`:

```
interval = interval.bothEndpoints(d => d.toUTC())
```

That's general-purpose enough to go into the Interval class and should still be convenient enough. We could also add `eachEndpoint` that takes two functions as arguments.

pietrovismara:
I like your idea better than mine but I'm not a big fan of the method name. 

`set` suggests a usage like `DateTime.set` where you pass an object, while `setBothEndpoints` works more like a `map` function, returning a new instance of `DateTime` for each one of those contained in the interval. 

Perhaps `setBothEndpoints` would work like this:

```
interval = interval.setBothEndpoints(() => { zone: 'UTC' })
```

While an hypothetical `mapBothEndpoints`:

```
interval = interval.mapBothEndpoints(d => d.toUTC())
```

Finally, since an interval contains by definition only 2 endpoints, we could remove `Both` from the name, as it should feel clear enough to have `set/mapEndpoints` for anyone who has read the `Interval` reference.

icambron:
I agree; `mapEndpoints` would be a good name

pietrovismara:
If that's it I can try to provide a PR in the next days if it's welcome. First time I put my hands in `luxon`'s code so I might need suggestion, but hopefully not as the implementation should be easy.

icambron:
@pietrovismara Go for it. Should be pretty straightforward.

PR #400, opened 2019-01-08
mapEndpoints implementation + test + docs

pietrovismara:
As discussed in https://github.com/moment/luxon/issues/386, here's my first try!


icambron:
Looks good, thanks!

PR #410, opened 2019-01-20
Fix links in moment.md

MilanMisak:


icambron:
Please see the docs readme here: https://github.com/moment/luxon/blob/master/docs/readme.md. I'd love to be able to make the same links work in both the markdown and the site, but I don't currently know of a way to do that.

Issue #417, opened 2019-01-26
add .js extension to import statements


m0ose:
I get an error on certain webservers when trying to import luxon using the es6 import statement. Some webservers are configured to add the `.js` file extension but It is the exception. 
```GET ../node_modules/luxon/src/datetime net::ERR_ABORTED 404 (Not Found)```
Seriously, does dropping the extension really save that much space or make the code more readable?
I dont mind making a pull request if it is ok.

icambron:
Fixed in #419

PR #419, opened 2019-01-28
added .js to imports

m0ose:
Put the file extension into import statements for all of the files in the src directory. This enables Luxon to be imported as a module by browsers that support the import statement. It currently works with some servers that ad the extension but not all. This should offer much faster download speeds over the HTTP/2 protocol. 
Address the issue #417


43081j:
Browsers and run-times like deno need extensions for module resolution (as it is done over HTTP), so +1 on this.

Much needed so we can use it as-is.

icambron:
OK, sure.

PR #420, opened 2019-01-29
fix zero duration formatting invalid value

tai2:
from [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601)

>Date and time elements including their designator may be omitted if their value is zero, and lower order elements may also be omitted for reduced precision. For example, "P23DT23H" and "P4Y" are both acceptable duration representations. **However, at least one element must be present, thus "P" is not a valid representation for a duration of 0 seconds. "PT0S" or "P0D", however, are both valid and represent the same duration.**

This PR makes it conform this description.


icambron:
Thanks!

Issue #402, opened 2019-01-09
Passing string values in durationObject multiplies it on 10.


franzzua:
line 3 should be same as line 2 or throw error
```
const {DateTime} = require('luxon');

const dateTime = DateTime.utc(2019,1,1);

console.log(dateTime.toISO());                            // 2019-01-01T00:00:00.000Z
console.log(dateTime.plus({ minutes: 1 }).toISO());       // 2019-01-01T00:01:00.000Z
console.log(dateTime.plus({ minutes: "1" }).toISO());     // 2019-01-01T00:10:00.000Z
```
Same thing with { days: "1" } or { seconds: "1" }

ricardobanegas:
Noticed bug also, I took care of it by parsing value to int.

icambron:
I'd take a PR that throws on a non-numeric value. I bet it goes somewhere in this Utils.normalizeObject: https://github.com/moment/luxon/blob/master/src/impl/util.js#L192

icambron:
Fixed in #421

PR #421, opened 2019-01-29
handled string values while normalizeObject

waseemahmad31:
Fixed issue #402 

line 3 should be same as line 2 or throw error
```javascript
const {DateTime} = require('luxon');

const dateTime = DateTime.utc(2019,1,1);

console.log(dateTime.toISO());                            // 2019-01-01T00:00:00.000Z
console.log(dateTime.plus({ minutes: 1 }).toISO());       // 2019-01-01T00:01:00.000Z
console.log(dateTime.plus({ minutes: "1" }).toISO());     // 2019-01-01T00:10:00.000Z
```
Same thing with { days: "1" } or { seconds: "1" }

`The issue was because we were performing the math operation without converting values in the number 
So if we add "1" + 0 in js it will be "10". `

### DIFF
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -194,10 +194,11 @@ export function normalizeObject(obj, normalizer, ignoreUnknown = false) {
   for (const u in obj) {
     if (obj.hasOwnProperty(u)) {
       const v = obj[u];
-      if (v !== null && !isUndefined(v) && !Number.isNaN(v)) {
+      const numericValue = Number(v);
+      if (v !== null && !Number.isNaN(numericValue)) {
         const mapped = normalizer(u, ignoreUnknown);
         if (mapped) {
-          normalized[mapped] = v;
+          normalized[mapped] = numericValue;
         }
       }
     }
--- a/test/datetime/create.test.js
+++ b/test/datetime/create.test.js
@@ -574,10 +574,10 @@ test("DateTime.fromObject handles null as a language tag", () => {
 });
 
 test("DateTime.fromObject overrides invalid date part with the current date part", () => {
-  const dt = DateTime.fromObject({ months: 9, day: 13, year: "hello" });
+  const dt = DateTime.fromObject({ year: "hello" });
   const localDt = DateTime.local();
   expect(dt.isValid).toBe(true);
   expect(dt.year).toBe(localDt.year);
-  expect(dt.months).toBe(9);
-  expect(dt.day).toBe(13);
+  expect(dt.months).toBe(localDt.months);
+  expect(dt.day).toBe(localDt.day);
 });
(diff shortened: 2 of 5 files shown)


src/impl/util.js
```diff
@@ -197,7 +197,7 @@ export function normalizeObject(obj, normalizer, ignoreUnknown = false) {
       if (v !== null && !isUndefined(v) && !Number.isNaN(v)) {
         const mapped = normalizer(u, ignoreUnknown);
         if (mapped) {
-          normalized[mapped] = v;
+          normalized[mapped] = Number(v);
```
icambron: This isn't quite right, because it lets NaN into this. You need to convert to a Number above that `if` check a few lines above.

waseemahmad31: Thanks, @icambron for the review.

src/impl/util.js
```diff
@@ -194,10 +194,11 @@ export function normalizeObject(obj, normalizer, ignoreUnknown = false) {
   for (const u in obj) {
     if (obj.hasOwnProperty(u)) {
       const v = obj[u];
-      if (v !== null && !isUndefined(v) && !Number.isNaN(v)) {
+      const numericValue = Number(v);
+      if (v !== null && !Number.isNaN(numericValue)) {
```
waseemahmad31: @icambron, After changing this code, no invalid number string except (empty string or boolean value) pass the if condition. So if we test this function with this change for `DateTime.fromObject({ year: "hello" })` it will return the valid date-time with the value DateTime.Now because of the below line https://github.com/moment/luxon/blob/c16d7d14f6678886f50e3bb457c0b6e07f91470f/src/datetime.js#L646
Is it the correct behaviour??

Paul-Todd:
Why don't we just add a check Number.isInteger(v) to line 197? This would also mean we can remove the null, undefined and NAN checks as well, or did I miss something?

```
if (v !== null && !isUndefined(v) && !Number.isNaN(v))
```
to
```
if (Number.isInteger(v)) {
....
}
```

waseemahmad31:
> Why don't we just add a check Number.isInteger(v) to line 197? This would also mean we can remove the null, undefined and NAN checks as well, or did I miss something?
> 
> ```
> if (v !== null && !isUndefined(v) && !Number.isNaN(v))
> ```
> to
> 
> ```
> if (Number.isInteger(v)) {
> ....
> }
> ```

@Paul-Todd thanks for the suggession but Luxon supports the fractional part also, so we can not use Number.isInteger(v).

src/impl/util.js
icambron: I think that is the correct behavior, yeah

icambron:
@waseemahmad31 So that "hello" test just needs to change. Then we can merge this.

waseemahmad31:
> @waseemahmad31 So that "hello" test just needs to change. Then we can merge this.

Sure, I will change it

icambron:
thanks! I will release this soon

Issue #377, opened 2018-12-06
`toLocal` should respect `Settings.defaultTimeZone`


abhikp:
For testing purposes, setting the zone for both `local()` and `toLocal()` is helpful. An example would be in-browser tests or showing React component state that's dependent on time in StorybookJS.

icambron:
This was fixed in #423

PR #423, opened 2019-01-30
changes made for toLocal to accept the default Zone

waseemahmad31:
Fixed issue #377 

toLocal will accepts `Settings.defaultZone` as default timezone if it is defined.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1226,7 +1226,7 @@ export default class DateTime {
    * @return {DateTime}
    */
   toLocal() {
-    return this.setZone(new LocalZone());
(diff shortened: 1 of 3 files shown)


icambron:
This is failing a test because of an interaction with another bug. I'm going to do a little refactoring here and also fix up the bug that's causing the test to fail. Then I'll merge the whole lot of it.

icambron:
Merged in 157af9395d4c0e1cf42e21735bbf43d994b703b3. Thanks for the contribution!

abhikp:
Thanks for changing this!

PR #424, opened 2019-01-30
Make matrix reflect MDN compatibility table

cmitz:
Hi, I was impressed that FF had no full Intl support, so I checked. It came in FF 65*, so I updated the table.

*According to MDN docs: [compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Browser_compatibility)

p.s. I'm not sure this was the right way to do this, since you mention in the contribution guide that clarifications and gotchas should be in the docstrings instead of the guides on the docs page. Feel free to close if this is not what you want.


icambron:
Thanks!

Issue #407, opened 2019-01-14
toRelativeCalendar shows today twice when using setZone


ramyma:
When a date with **setZone** is used with **toRelativeCalendar** function, _today_ is returned for today and for the day following (tomorrow), then it shows _tomorrow_ and starts counting for subsequent days.

You can check this example https://codesandbox.io/s/7z2j4552zx to see it in action.

```
import { DateTime } from "luxon";

const date = DateTime.local().setZone(`UTC+3`);
const date2 = date.plus({ days: 1 });
const date3 = date2.plus({ days: 1 });
const date4 = date3.plus({ days: 1 });

console.log(date.toRelativeCalendar());
console.log(date2.toRelativeCalendar());
console.log(date3.toRelativeCalendar());
console.log(date4.toRelativeCalendar());
```

Resolves to:
```
today         ->    2019-01-14T04:24:29.834+03:00 
today         ->    2019-01-15T04:24:29.834+03:00 
tomorrow      ->    2019-01-16T04:24:29.834+03:00 
in 2 days     ->    2019-01-17T04:24:29.834+03:00 
```

Notice how it shows _today_ twice for two different days.

Did anyone run into this issue ?

Version: **1.10.0**

icambron:
Fixed by #430

PR #430, opened 2019-02-06
fixed issue "toRelativeCalendar shows today twice when using setZone"

waseemahmad31:
Fixed issue #407

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1763,7 +1763,7 @@ export default class DateTime {
    */
   toRelative(options = {}) {
     if (!this.isValid) return null;
-    const base = options.base || DateTime.local().setZone(this.zone.name),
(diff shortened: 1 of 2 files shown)


src/datetime.js
```diff
@@ -1763,7 +1763,7 @@ export default class DateTime {
    */
   toRelative(options = {}) {
     if (!this.isValid) return null;
-    const base = options.base || DateTime.local(),
+    const base = options.base || DateTime.local().setZone(this.zone.name),
```
icambron: I'd prefer this was `DateTime.fromObject({ zone: this.zone })`

waseemahmad31: @icambron  Thanks for the suggestion

icambron:
Thanks!

Issue #408, opened 2019-01-14
DateTime.fromFormat throws exception if used with IANA time zones


DanielSWolf:
The `DateTime` methods `toFormat()` and `fromFormat` are symmetric. That is, I can use `toFormat()` to convert a `DateTime` to a string value; then I can use the same format string to read the string value back into a `DateTime`.

That works well for most time-zone formats from [the documented table of tokens](https://moment.github.io/luxon/docs/manual/formatting.html#table-of-tokens). But when using `z` for IANA zones, `fromFormat()` throws an exception.

Here's a demo:

```js
const { DateTime } = require('luxon');

const d = DateTime.fromObject({ year: 2019, month: 01, day: 14, hour: 11, minute: 30, zone: 'Indian/Maldives' });

const shortOffsetFormat = "yyyy-MM-dd'T'HH-mmZZ";
// 🙂 Prints "2019-01-14T11-30+05:00"
console.log(d.toFormat(shortOffsetFormat));
// 🙂 Prints 5
console.log(DateTime.fromFormat('2019-01-14T11-30+05:00', shortOffsetFormat, { setZone: true }).zone.offset(d) / 60);

const ianaFormat = "yyyy-MM-dd'T'HH-mm[z]";
// 🙂 Prints "2019-01-14T11-30[Indian/Maldives]"
console.log(d.toFormat(ianaFormat));
// 😥 Should print "Indian/Maldives", but throws exception "Invalid regular expression: /[/: Unterminated character class"
console.log(DateTime.fromFormat('2019-01-14T11-30[Indian/Maldives]', ianaFormat, { setZone: true }).zoneName);
```

Here is the full exception:

```text
SyntaxError: Invalid regular expression: /[/: Unterminated character class
    at RegExp (<anonymous>)
    at literal (/Users/wolf/dev/luxon-test/node_modules/luxon/build/node/luxon.js:4081:12)
    at unitate (/Users/wolf/dev/luxon-test/node_modules/luxon/build/node/luxon.js:4087:14)
    at unitForToken (/Users/wolf/dev/luxon-test/node_modules/luxon/build/node/luxon.js:4240:16)
    at tokens.map.t (/Users/wolf/dev/luxon-test/node_modules/luxon/build/node/luxon.js:4366:33)
    at Array.map (<anonymous>)
    at explainFromTokens (/Users/wolf/dev/luxon-test/node_modules/luxon/build/node/luxon.js:4366:24)
[...]

PR #432, opened 2019-02-09
fixed special character issue in format string

waseemahmad31:
Fixes issue #408 

```javascript
// 😥 Should print "Indian/Maldives", but throws exception "Invalid regular expression:
// /[/: Unterminated character class"
const { DateTime } = require('luxon');
console.log(DateTime.fromFormat('2019-01-14T11-30[Indian/Maldives]', ianaFormat, 
{ 
   setZone: true 
}).zoneName);

```

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -38,6 +38,11 @@ function simple(regex) {
   return { regex, deser: ([s]) => s };
 }
 
+function escapeToken(value) {
+  // eslint-disable-next-line no-useless-escape
+  return value.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&");
+}
+
 function unitForToken(token, loc) {
   const one = /\d/,
     two = /\d{2}/,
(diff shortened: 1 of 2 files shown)


icambron:
This is a clever fix, but it doesn't quite work. Two problems:

First, this fails the same way as the original issue, because you're only escaping the first character:

```js
> DateTime.fromFormat("2019-01-14T11-30[[Indian/Maldives]]", "yyyy-MM-dd'T'HH-mm[[z]]")
SyntaxError: Invalid regular expression: /\[[/: Unterminated character class
```

The second issue is that `\` does more than escape; it also encodes literals like tab (`\t`), so you get this behavior:

```js
> DateTime.fromFormat("2019-01-14T11-30tIndian/Maldivest", "yyyy-MM-dd'T'HH-mm't'z't'").isValid
false
> DateTime.fromFormat("2019-01-14T11-30\tIndian/Maldives\t", "yyyy-MM-dd'T'HH-mm't'z't'").isValid
true
```

which is backwards. I think you might need to really go through the characters and escape them individually, which is a bummer.

waseemahmad31:
Ahhhh, did not thought this way. I will try again to fix this issue.

waseemahmad31:
used regex `/[\-\[\]{}()*+?.,\\\^$|#\s]/g` to escape special characters (for regular expression). Have a look and let me know if it works.

icambron:
Yes, looks great!

PR #436, opened 2019-02-11
Add Math.sign polyfill for IE

### LINKED ISSUE #434: Add Math.sign to IE11 Polyfill specifier
Luxon uses `Math.sign` now which has no IE11 support. This breaks a lot of functionality, as can been seen running the `demo` in IE11. Simply fix is to add `Math.sign` to the `polyfill.io` callout and `demo`'s global link.

It's a simple change, so I can PR if yah'd like :)

RichiCoder1:
Fixes #434.

Adds `Math.sign` to the polyfill.io url for the global demo, and the IE install docs.


icambron:
Thanks!

Issue #448, opened 2019-02-22
DateTime.fromObject silently accepts any wrong value


maxp:
That behaviour is completely wrong and dangerous
```
> luxon.DateTime.fromObject({year:"wrong value"}).isValid
true
```

icambron:
I'd accept a pull request that threw here

icambron:
Fixed in #452

PR #451, opened 2019-02-25
DateTime.fromObject silently accepts any wrong value #448

GillesDebunne:
Make the `fromObject()` and `set` methods on `DateTime` and `Duration` more strict.

Invalid keys in the object parameter, as well as invalid values for units (such as `NaN`, `true` or `""`) now throw an error (or create an invalid `DateTime`). See the updated tests for details.

Various other minor changes.


GillesDebunne:
Fixing broken test and re-submitting

PR #452, opened 2019-02-25
DateTime.fromObject silently accepts any wrong value #448

GillesDebunne:
Make the `fromObject()` and `set` methods on `DateTime` and `Duration` more strict.

Invalid keys in the object parameter, as well as invalid values for units (such as `NaN`, `true` or `""`) now throw an error (or create an invalid `DateTime`). See the updated tests for details.

Various other minor changes.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -155,12 +155,13 @@ function parseDataToDateTime(parsed, parsedZone, opts, format, text) {
   const { setZone, zone } = opts;
   if (parsed && Object.keys(parsed).length !== 0) {
     const interpretationZone = parsedZone || zone,
-      obj = Object.assign(parsed, opts, {
-        zone: interpretationZone
-      });
-    // setZone is a valid option in calling methods, but not in fromObject
-    delete obj.setZone;
-    const inst = DateTime.fromObject(obj);
+      inst = DateTime.fromObject(
+        Object.assign(parsed, opts, {
+          zone: interpretationZone,
+          // setZone is a valid option in the calling methods, but not in fromObject
+          setZone: undefined
+        })
+      );
     return setZone ? inst : inst.setZone(zone);
   } else {
     return DateTime.invalid(
@@ -590,20 +591,14 @@ export default class DateTime {
       return DateTime.invalid(unsupportedZone(zoneToUse));
     }
 
-    let normalized;
-    try {
+    const tsNow = Settings.now(),
+      offsetProvis = zoneToUse.offset(tsNow),
       normalized = normalizeObject(obj, normalizeUnit, [
         "zone",
         "locale",
         "outputCalendar",
         "numberingSystem"
-      ]);
-    } catch (error) {
-      return DateTime.invalid(error.message);
-    }
-
-    const tsNow = Settings.now(),
-      offsetProvis = zoneToUse.offset(tsNow),
+      ]),
       containsOrdinal = !isUndefined(normalized.ordinal),
       containsGregorYear = !isUndefined(normalized.year),
       containsGregorMD = !isUndefined(normalized.month) || !isUndefined(normalized.day),
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -155,11 +155,12 @@ function parseDataToDateTime(parsed, parsedZone, opts, format, text) {
   const { setZone, zone } = opts;
   if (parsed && Object.keys(parsed).length !== 0) {
     const interpretationZone = parsedZone || zone,
-      inst = DateTime.fromObject(
-        Object.assign(parsed, opts, {
-          zone: interpretationZone
-        })
-      );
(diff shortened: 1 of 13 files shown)


src/impl/util.js
```diff
@@ -208,18 +216,20 @@ export function signedOffset(offHourStr, offMinuteStr) {
 
 // COERCION
 
-export function normalizeObject(obj, normalizer, ignoreUnknown = false) {
+function asNumber(value) {
+  const numericValue = Number(value);
+  if (typeof value === "boolean" || value === "" || Number.isNaN(numericValue))
+    throw new Error(`Invalid unit value ${value}`);
```
GillesDebunne: Auto comment here. This could be a Luxon Error subclass, such as `InvalidArgumentError`. But this would add a dependency to `errors.js` in what is today a leaf in the include dependencies.

src/datetime.js
```diff
@@ -155,11 +155,12 @@ function parseDataToDateTime(parsed, parsedZone, opts, format, text) {
...
-      inst = DateTime.fromObject(
-        Object.assign(parsed, opts, {
-          zone: interpretationZone
-        })
-      );
+      obj = Object.assign(parsed, opts, {
+        zone: interpretationZone
+      });
+    // setZone is a valid option in calling methods, but not in fromObject
+    delete obj.setZone;
```
icambron: would be a little nicer to just put `setZone: null` in the `assign` call just above

src/datetime.js
```diff
@@ -589,9 +590,20 @@ export default class DateTime {
...
+    let normalized;
+    try {
+      normalized = normalizeObject(obj, normalizeUnit, [
+        "zone",
+        "locale",
+        "outputCalendar",
+        "numberingSystem"
+      ]);
+    } catch (error) {
+      return DateTime.invalid(error.message);
```
icambron: I'm not sure about this. The rule in Luxon is that we want to throw on programmer errors and invalidate on data errors. That roughly translates to bad keys and bad types should throw, out-of-range numbers or unknown strings should invalidate. So I think here we should probably just throw all the way up.

src/impl/util.js
```diff
@@ -88,9 +88,17 @@ export function padStart(input, n = 2) {
...
+  } else {
+    return parseInt(string, 10);
+  }
+}
+
 export function parseMillis(fraction) {
-  if (isUndefined(fraction)) {
-    return NaN;
+  if (isUndefined(fraction) || fraction === null || fraction === "") {
+    return undefined;
```
icambron: Why this?

src/impl/util.js
icambron: I hear you, but I think we should be consistent, even it creates some dependency cycles.

src/impl/util.js
```diff
@@ -208,18 +216,20 @@ export function signedOffset(offHourStr, offMinuteStr) {
...
     if (obj.hasOwnProperty(u)) {
       const v = obj[u];
-      const numericValue = Number(v);
-      if (v !== null && !Number.isNaN(numericValue)) {
-        const mapped = normalizer(u, ignoreUnknown);
-        if (mapped) {
-          normalized[mapped] = numericValue;
-        }
-      }
+      if (v === undefined || v === null || nonUnitKeys.includes(u)) continue;
```
icambron: I think you meant to check on `u`, not `v`?

Also, the polyfill build/instructions would need to add `includes`. If we go through all of our `indexOf(...) !== -1`s and change them, that's worth it, but as it stands, let's just use indexOf here

icambron:
Lot of good changes in here. Left some comments

src/impl/util.js
GillesDebunne: I agree.

src/impl/util.js
GillesDebunne: Testing for `null`, `undefined` and "" since all three would otherwise result in a 0 output on the next line. Semantically, milliseconds are not set in these cases, so 0 is not an actual representation (although it would probably break no tests).

Using `undefined` because it will be ignored, while `NaN` would now be considered an invalid value.

I'll add a comment.

src/datetime.js
GillesDebunne: Done. That was not an option with a previous version of my code.

src/datetime.js
GillesDebunne: Makes sense. I thought the idea was to avoid throwing and use invalid with a reason instead.

But that would be inconsistent with Duration.

src/impl/util.js
GillesDebunne: `null` and `undefined` *values* are indeed ignored, since both would be converted to number 0 in `normalized`.

Switched to `indexOf`.

icambron:
I'm going to go ahead and merge this. Thanks for the contribution!

PR #453, opened 2019-02-28
#448 invalid numerical value check

maxp:
Sorry for that dirty code - unable to run tests because of babel version incompatibility :(


icambron:
Hi @maxp, I ended up taking #451 for this fix, since it covered a bit more. Thanks for the contribution, though!

PR #456, opened 2019-03-02
Return defaultZone from normalizeZone if input is local

### LINKED ISSUE #454: DateTime not using `defaultZoneName` set via `Settings`
I'm trying to write Node/Jest tests for an implementation and it looks like I am not able to override the default timezone as described in the docs [here](https://github.com/moment/luxon/blob/master/docs/zones.md#changing-the-default-zone).  The implementation itself (React) works fine but something is not sticking on the node side; DateTime is correctly reporting the new `zoneName` but is not formatting with that zone.  

This is what I'm seeing in the node shell (my timezone is Berlin/GMT+1):

```js
$ node
> const { Settings, DateTime } = require('luxon')
undefined
> DateTime.local().zoneName
'Europe/Berlin'
> DateTime.fromSeconds(1541590322, { zone: 'utc' }).setZone('local').toFormat('EEE, dd MMM y, T');
'Wed, 07 Nov 2018, 12:32'
> Settings.defaultZoneName = "America/Los_Angeles"
'America/Los_Angeles'
> DateTime.local().zoneName
'America/Los_Angeles'
> DateTime.fromSeconds(1541590322, { zone: 'utc' }).setZone('local').toFormat('EEE, dd MMM y, T');
'Wed, 07 Nov 2018, 12:32'
> DateTime.local().zoneName
'America/Los_Angeles'
```

My understanding is that this is fixed by https://github.com/moment/luxon/pull/423 (`Settings.defaultZone` is reporting the Los Angeles zone as expected also)

I've tried this with both node v8.11.4 and 10.15.1

Thank you in advance, I'm enjoying using this lib over Moment :)

mvpowers:
Fixes #454

### DIFF
--- a/src/impl/zoneUtil.js
+++ b/src/impl/zoneUtil.js
@@ -3,7 +3,6 @@
  */
 
 import Zone from "../zone.js";
-import LocalZone from "../zones/localZone.js";
 import IANAZone from "../zones/IANAZone.js";
 import FixedOffsetZone from "../zones/fixedOffsetZone.js";
 import InvalidZone from "../zones/invalidZone.js";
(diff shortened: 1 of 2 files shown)


icambron:
This looks right, but can you add a test?

mvpowers:
Sure thing! @icambron Test added

icambron:
Merged in 5b9e59d8ccfd39d2e7fa8716952967ad50b90773. Thanks for the contribution!

PR #458, opened 2019-03-04
Add tests and minor fixes, so that DateTime code coverage reaches 100%

GillesDebunne:


src/datetime.js
```diff
@@ -715,7 +714,7 @@ export default class DateTime {
    * @param {string} opts.outputCalendar - the output calendar to set on the resulting DateTime instance
    * @param {string} opts.numberingSystem - the numbering system to set on the resulting DateTime instance
    * @example DateTime.fromRFC2822('25 Nov 2016 13:23:12 GMT')
-   * @example DateTime.fromRFC2822('Tue, 25 Nov 2016 13:23:12 +0600')
+   * @example DateTime.fromRFC2822('Fri, 27 Apr 2018 09:27:12 +0200')
```
icambron: Why?

src/datetime.js
```diff
@@ -1211,7 +1210,7 @@ export default class DateTime {
    * @return {DateTime}
    */
   toLocal() {
-    return this.setZone(Settings.defaultZone || new LocalZone());
```
icambron: Ha, I was just thinking about how I should make this change.

test/interval/create.test.js
```diff
@@ -4,7 +4,7 @@ import { DateTime, Interval, Duration, Settings } from "../../src/luxon";
 //------
 // .fromObject()
 //-------
-test("Interval.fromDateTimes creates an interval from datetimes", () => {
+test("Interval#fromDateTimes creates an interval from datetimes", () => {
```
icambron: hmm, no: `#` is for instance methods and `.` is for class (i.e. static) members

icambron:
Looks good. A few scattered comments

src/datetime.js
GillesDebunne: Because '25 Nov 2016' is not a Tuesday. But I'll fix it without changing the date.

icambron: Ah, no, that is a good reason. I don't care what the date is.

icambron:
Thanks! This is great

PR #463, opened 2019-03-06
Duration 100% coverage, prettier and jest configuration

GillesDebunne:
- Bump Duration code coverage to 100%
- Move prettier configuration to default config file
- Remove jest-cli and use jest-config.js so that jest can be run directly


.eslintrc.js
```diff
@@ -1,16 +1,14 @@
-const prettierOptions = require("./.prettier.js");
-
 module.exports = {
-  extends: ["standard", "prettier", "prettier/standard"],
+  extends: ["standard", "plugin:prettier/recommended"],
```
icambron: so does this automatically pull in the prettierrc?

GillesDebunne: As I understand it, `eslint-plugin-prettier` does report prettier problems as eslint errors, based on the `.prettierrc` config file.

While `eslint-config-prettier` turn off all other rules that would conflict with prettier. 

Using `recommended` does set `prettier/prettier` rule to "error" (i.e. it can be omitted), and it extends `config` with the actual `.prettierrc` configuration.

icambron:
Thanks!

benmccann:
@GillesDebunne @christianbender it looks to me like there was a bad merge between this commit and https://github.com/moment/luxon/commit/7b383aa1b75d494b158c0dd1096d40b253ceff53 because the building is now failing

GillesDebunne:
Yes, there is a prettier error. It is fixed in my PR.

I could submit a separate PR to fix this very error, which may be cleaner

PR #464, opened 2019-03-08
Update default locale documentation in factories

GillesDebunne:
Defaults to system's locale instead of `null` or en-US as explained in the doc.


icambron:
Merged in  af3b79647da4245c169f1d96240c53365441c476

icambron:
And thanks!

PR #470, opened 2019-03-15
Adds week token to Duration's ISO formatting

stefanpearson:
From the look of the source, it looks like converting weeks => days in formatting was intentional. However, I would expect `Duration.fromISO('P1W').toISO()` to be deterministic and preserved as `P1W` and not `P7D` 🙂.

#### Use case

This was causing problems for me when I was taking a saved ISO duration string and deserialising it into a UI that would allow you to build a duration. When a user sets [3] [Weeks] in a form, the next time it renders, it would appear as [21] [Days].

**NB:** I ❤️ Luxon

### DIFF
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -2,8 +2,7 @@
 /* eslint no-global-assign: "off" */
 import { DateTime, Settings } from "../src/luxon";
 
-
-exports.withoutIntl = function (name, f) {
+exports.withoutIntl = function(name, f) {
   const fullName = `With no Intl support, ${name}`;
   test(fullName, () => {
     const intl = Intl;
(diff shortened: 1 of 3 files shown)


Jafferwaffer:
Agreed with this change. As the ISO describes: `W is the week designator that follows the value for the number of weeks`.

test/helpers.js
```diff
@@ -2,8 +2,7 @@
 /* eslint no-global-assign: "off" */
 import { DateTime, Settings } from "../src/luxon";
 
-
-exports.withoutIntl = function (name, f) {
+exports.withoutIntl = function(name, f) {
```
stefanpearson: These changes were unintentional and must have been the linter? I can remove if necessary

icambron:
Thanks, great addition

Issue #469, opened 2019-03-15
1.11.4 fix which returns default timezone if set for the alias 'local' breaks everything


cravindra:
Hello.

Until the recent update to `1.11.4` using `local` time zone as a string explicitly set the zone to the browser default. While this was unexpected initially, we worked around it as follows:

```js
// Set default time zone for project:
Settings.defaultZoneName = timezone || 'local';

// To interpret server side dates given in ISO:
const zonedDate = DateTime.fromISO(date); // Interprets it as the project timezone

// Show on the UI: Must be scoped to browser timezone so as to show it in terms of the local time
// which may be different from the project timezone (Ex: A user in Japan selecting shift timings in a 
// date input for a project in India - The user in Japan must be able to select times (say 8am to 3pm) 
// in his/her worldview without having to do conversions
const displayDate = zoneDate.setZone('local', {keepLocalTime: true});

// Parse input from the date-picker in browser zone, convert to project timezone, convert to ISO for server
const utc = DateTime
                                .fromJSDate(date, {zone: 'local'})
                                 // Get project time zone by checking the zone of .a new DateTime object
                                .setZone(DateTime.local().zoneName, {keepLocalTime: true})
                                .setZone('utc')
                                .toMillis()

```

Now obviously this has stopped working. I have to admit that the new behaviour should be the expected behaviour. I was initially surprised by the older behaviour and had to code around it. Now this fix is breaking all our code 😆 . While this is something which can be fixed on our end, I feel there is a legitimate need to be able to quick access the browser default.

Perhaps we can introduce a new alias like `local` such as `system`? Or is there a better solution I'm not seeing? (I'd like to avoid the use of the timeZoneOffset from the native date object although that is one feasible solution)

icambron:
Yeah, I think that's a reasonable request. I hadn't even thought of this use case. I'd take a PR aliasing "system" to the local instance.

The workaround is to explicitly import the `LocalZone` object and pass `LocalZone.instance` to the `setZone` call.

 Also, just for completeness, I'll point out that you could also *not* set the default zone and just do:

 ```js
DateTime.fromISO(s, { zone: projectTimeZone }).toUTC()
```

cravindra:
Raised a PR: https://github.com/moment/luxon/pull/475

Using `LocalZone.instance` is a good idea.. I was actually unaware of that class until I dug into the code.

[...]

PR #475, opened 2019-03-21
Add `system` alias to LocalZone in normalizeZone

cravindra:
Add a `system` zone name alias to allow the system's default zone (LocalZone) to be used. This is required when the Luxon default zone has been modified and using `local` returns the default zone instead of the system zone.

Fixes: https://github.com/moment/luxon/issues/469


icambron:
Looks good, thanks!

icambron:
I ended up reverting this; see the discussion in #469. Sorry for telling you I'd accept this PR; I should have thought harder before giving the green light.

PR #485, opened 2019-04-01
Documents that parsing can fail, and should be checked.

marianogappa:
Parsing `DateTime`s fails silently, and with no documentation of this and no types it's impossible to know. This PR adds a comment to the return JSDoc. Please merge this PR and retrigger the doc generation on the site.


icambron:
There's an entire page of documentation about it: https://moment.github.io/luxon/docs/manual/validity.html. I agree that individual calls should describe under what conditions they will return invalid instances for, but it should go in the description and be specific, not the return type docs.

marianogappa:
@icambron https://github.com/moment/luxon/pull/490

PR #486, opened 2019-04-01
simplified token parser for the "z" zone token. added relevant fromFormatExplain test for the same. also fixed a failing test in create.test.js

shripadk:
Parses the zone token correctly. The current regex creates extra substring matches causing it to fail if the zone token is placed anywhere other than at the end of the format string. Here is an example:

Screenshot 2019-04-01 at 6 00 20 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "luxon",
-  "version": "1.11.4",
+  "version": "1.12.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


test/datetime/create.test.js
```diff
@@ -571,7 +571,7 @@ test("DateTime.fromObject accepts a locale string with weird junk in it", () =>
     });
 
     expect(res.locale).toBe("be");
-    expect(res.outputCalendar).toBe("gregory");
+    expect(res.outputCalendar).toBe("coptic");
```
shripadk: On my system (macOS 10.14.3) the test fails with `outputCalendar` expecting **coptic** instead of **gregory**. Had no choice but to fix the test before committing.

More info:

```bash
 ~/@/l/luxon $ node -v
v11.13.0
```
Node compiled and installed with-full-icu support.

icambron:
The failing create test is a node 11 vs 10 thing. I have a plan to handle that, but in the meantime can you revert the change?

shripadk:
@icambron I downgraded to Node 10 but got the same error. I think it probably has to do with macOS 10.14.3. I tested it in Chrome (navigating to Luxon docs and trying it out in the console) and got "coptic" instead of "gregory":

Screenshot 2019-04-01 at 7 18 17 PM

Will try with your docker image (icambron/luxon) and report back if I get a different result. For now I have reverted those changes as it was clearly failing the Travis build.

icambron:
Interesting. That will complicate my plan to fix that test.

I will look at the PR content itself soon

icambron:
Yes, this makes sense.

icambron:
Merged in 4326774632fef5c0da4a7e8cb8af8bd089d6f8ab. Thanks!

shripadk:
Thank you! 👍

PR #488, opened 2019-04-03
Localized fromFormatExplain with numberingSystem support

shripadk:
**Problem**: Running `fromFormatExplain` on a localized string with a numberingSystem defined would have `rawMatches` return `null` and `matches` an empty object:

Screenshot 2019-04-03 at 1 15 25 PM

With this PR, it is now possible to parse localized numberingSystem strings correctly:

Screenshot 2019-04-03 at 1 20 04 PM

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -157,7 +157,7 @@ function unitForToken(token, loc) {
         // we don't support ZZZZ (PST) or ZZZZZ (Pacific Standard Time) in parsing
         // because we don't have any way to figure out what they are
         case "z":
-          return simple(/[a-z_+-]{1,256}(\/[a-z_+-]{1,256}(\/[a-z_+-]{1,256})?)?/i);
+          return simple(/[a-z_+-/]{1,256}?/i);
         default:
           return literal(t);
       }
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -2,11 +2,12 @@ import { parseMillis, isUndefined, untruncateYear, signedOffset } from "./util.j
 import Formatter from "./formatter.js";
 import FixedOffsetZone from "../zones/fixedOffsetZone.js";
 import IANAZone from "../zones/IANAZone.js";
+import { digitRegex, parseDigits } from "./digits.js";
 
 const MISSING_FTP = "missing Intl.DateTimeFormat.formatToParts support";
 
 function intUnit(regex, post = i => i) {
-  return { regex, deser: ([s]) => post(parseInt(s, 10)) };
+  return { regex, deser: ([s]) => post(parseDigits(s)) };
 }
 
 function fixListRegex(s) {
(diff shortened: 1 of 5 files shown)


src/impl/tokenParser.js
```diff
@@ -64,32 +69,32 @@ function unitForToken(token, loc) {
...
         case "yyyyyy":
-          return intUnit(/\d{6}/);
+          return intUnit(six);
         // months
         case "M":
           return intUnit(oneOrTwo);
         case "MM":
           return intUnit(two);
         case "MMM":
-          return oneOf(loc.months("short", false, false), 1);
```
icambron: I don't see why these should change from false to true?

icambron:
This totally makes sense and I'm not really sure how to improve the code from what you have. I'm just a little bummed that it's so much code. Let me think on this one a bit

src/impl/tokenParser.js
shripadk: It is to bring parity with format string in `toFormat`. For instance, let us consider the locale to be `zh` and numberingSystem to be `hanidec`. It would be something like this with the above fix:

Screenshot 2019-04-06 at 10 31 23 AM

Earlier, `loc.months` returned an array of months in a different format which made it difficult to easily transpose the output of `toFormat` with `fromFormat`. The ideal situation would be that the user can feed the output of `toFormat` to `fromFormat` and get back the original value. I know that it doesn't work in some cases (like if the format has `EEEEE` making it ambiguous. does "T" stand for "Tuesday" or "Thursday"?).

From my tests, the output of:

`loc.months("long", false, false)` for `hanidec` numberingSystem was:

`[ '一月', '二月', '三月', '四月', '五月', '六月', '七月', '八月', '九月', '十月', '十一月', '十二月' ]`

As you can clearly see the `toFormat` just output the numerical equivalent of "April" as `四` but `loc.months` output the string equivalent of "April" as `四月`. The parser was unable to parse the string (because there is no `四月` in the localized string) causing it to output `null` for rawMatches. The correct output for `loc.months` should be:

`[ '一', '二', '三', '四', '五', '六', '七', '八', '九', '一〇', '一一', '一二' ]`

I think your intent was right but the flags were probably flipped in the implementation of `Locale.months` in locale.js. I was not sure about it so I did not touch the locale.js code. I instead flipped the boolean in the parser logic itself. It did not cause any other tests to fail so I am assuming it wasn't tested earlier and my tests were the first to check that line of code.

Nevertheless please do let me know if my understanding was incorrect. I am new to your library (which is absolutely amazing by the way) and am learning the internals on the go. 😃

shripadk:
@icambron Absolutely no issues! Please take your time and do let me know if you encounter something confusing in the code. Thanks!

shripadk:
Refactored `digits.js` file to only have a single source of truth for unicode endpoints for various numberingSystems. This should make maintenance easier if and when new numberingSystems are added by browser vendors.

Also added a test to ensure that future additions or changes always fall within the desired range (must map one-to-one with latin digits 0-9).

icambron:
I've decided I'm going to merge this

icambron:
Merged in cbc9e6e621241634fb03ee417af7a97000dd400f

PR #490, opened 2019-04-09
Documents that .fromISO can fail and what to do.

marianogappa:
Opened as a result of:
https://github.com/moment/luxon/pull/485


icambron:
I removed invalids in the 2.0 branch, which should come out in the next month or so. Since you haven't filled out the CLA and I can't merge without it, I'm going to close this.

Issue #494, opened 2019-04-24
Settings.throwOnInvalid does not throw for all invalid dates


JStonevalley:
Hi
I have found that setting throwOnInvalid does not result in an error being thrown for all dates that are invalid.

The following line results in an invalid date but does not throw:
```javascript
Settings.throwOnInvalid = true  
console.log(DateTime.fromJSDate('2019-04-16T11:32:32Z').isValid) // false
```

Full working example can be found at: https://runkit.com/embed/lr17m8g8q296

icambron:
yup, looks like a bug

icambron:
Fixed in #497

PR #497, opened 2019-05-03
throw error on invalid date if throwOnInvalid  is true

waseemahmad31:
fix issue : #494 

if Settings.throwOnInvalid is true, then DateTime.fromJSDate will throw error for invalid dates


icambron:
Looks good, thanks!

PR #500, opened 2019-05-07
combined regexes and extractors are created in advance to speed up fromISO() and fromSQL()

ntamas:
This PR tweaks the internal implementation of `DateTime.fromISO()` and `DateTime.fromSQL()` such that the regexes and extractors used during parsing are created once _in advance_ and not with every invocation of these functions. This resulted in a speedup of about +25% on my machine for both functions (from ~55,000 ops/sec to ~67,000 ops/sec).


icambron:
Very nice. Thanks!

icambron:
Release in 1.13.1

PR #507, opened 2019-05-09
Fixed comments/documentation for startOf function

awei01:
Was missing valid arguments `quarter` and `week` for `startOf` method


icambron:
Thanks

Issue #504, opened 2019-05-08
Feature Request: add a method to Zone returning `+05:00`


ycmjason:
`Zone` currently has a method `offsetName(ts, opts)` which returns `AST` or its long form.

I suggest having an equivalent function (or extend `offsetName`'s `opts`) which can retrieve the formatted offset `+05:00`.

icambron:
This would be a new method called `formatOffset(ts, format)`, where the `format` parameter with allowed values "short", "narrow", and "techie". The implementation should get moved from [here](https://github.com/moment/luxon/blob/236f2badea297ee73421aa39a83e06177c1be6d0/src/impl/formatter.js#L130) to Zone.js, documented, and of course called from the formatter.

icambron:
Fixed in #509

PR #509, opened 2019-05-15
Add formatOffset method to Zones.

mastermatt:
For feature request #504.

Moves the existing formatting logic from formatter.js to util.js so it can be utilized by all the zone subclasses.

@icambron I'd be happy to add docs and/or tests but I don't see any that deal with interacting with Zone instances directly, so I wasn't sure which files to update.

### DIFF
--- a/src/impl/formatter.js
+++ b/src/impl/formatter.js
@@ -132,21 +132,7 @@ export default class Formatter {
           return "Z";
         }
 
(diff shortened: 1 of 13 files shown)


icambron:
Yeah, we should have a new `test/zones` directory with tests in it.

mastermatt:
@icambron I've added those tests.

icambron:
Awesome, thanks

PR #517, opened 2019-05-28
Fix inconclusive error message

### LINKED ISSUE #516: Inconclusive error message: Duration.fromObject: argument expected to be an object, got object
If you pass `null` to `Duration.fromObject()` you'll get the following error message:

> Duration.fromObject: argument expected to be an object, got object

this happens in the following lines of code ([duration.js:272](https://github.com/moment/luxon/blob/master/src/duration.js#L227,L231)):

```
if (obj == null || typeof obj !== "object") {
      throw new InvalidArgumentError("Duration.fromObject: argument expected to be an object, got " + typeof obj);
}
```

The code is not incorrect but the error message inconclusive. This is because of the not using a strict comparison operator and using typeof which doesn't always have the desirable result.

Maybe this could be a bit improved…

shaedrich:
`null` is typeof "object"

Fix #516

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -226,7 +226,7 @@ export default class Duration {
   static fromObject(obj) {
     if (obj == null || typeof obj !== "object") {
       throw new InvalidArgumentError(
-        `Duration.fromObject: argument expected to be an object, got ${typeof obj}`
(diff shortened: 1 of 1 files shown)


icambron:
Looks good, thanks!

PR #524, opened 2019-06-16
Update formatting.md

corysimmons:
Fix link to API ref for `toFormat`

Note this isn't relative to wherever docs are. Might want to tweak it.

### DIFF
--- a/docs/formatting.md
+++ b/docs/formatting.md
@@ -122,7 +122,7 @@ You shouldn't create ad-hoc string formats if you can avoid it. If you intend fo
 
 ### toFormat
 
-See [DateTime#toFormat](../class/src/datetime.js~DateTime.html#instance-method-toFormat) for the API signature. As a brief motivating example:
+See [DateTime#toFormat](https://moment.github.io/luxon/docs/manual/formatting.html#table-of-tokens) for the API signature. As a brief motivating example:
 
 ```js
 DateTime.fromISO('2014-08-06T13:07:04.054').toFormat('yyyy LLL dd'); //=> '2014 Aug 06'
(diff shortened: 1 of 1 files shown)


docs/formatting.md
```diff
@@ -122,7 +122,7 @@ You shouldn't create ad-hoc string formats if you can avoid it. If you intend fo
 
 ### toFormat
 
-See [DateTime#toFormat](../class/src/datetime.js~DateTime.html#instance-method-toFormat) for the API signature. As a brief motivating example:
```
icambron: I don't understand this. The link is supposed to be to the method signature for `toFormat`, not to the table of tokens...

corysimmons: You're right. Changed to https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toFormat

icambron:
I just checked this on the website and it seems to work as expected. I think I see the misunderstanding here, now that I've looked closer. I think you're expecting the link to work in the markdown file in the repository? Please see https://github.com/moment/luxon/blob/master/docs/readme.md.
 
I guess there's a real question of how it should work and definitely it would be nice if you could follow the links in the docs markdown. But there are two issues:

 1. Links _between_ files in the markdown don't work at all now, because the links point to the website paths. That would need to be handled by some kind of link rewriter in the website publishing step.
 2. There are a _lot_ of links between those md files and the generated api docs. I don't want to set the expectation that they work when most of them don't.

So I've mostly concluded that for now, I'm only supporting these docs in their published-on-the-website form. I'm going to close this. Thanks for taking a crack at it, though; I recognize this trips a bunch of people up.

corysimmons:
Oh yeah, you're right. My bad. =)

Issue #521, opened 2019-06-04
If Interval is invalid due to Start Date coming after end date, it should throw if Settings.throwOnInvalid is true


DanielJoyce:
https://github.com/moment/luxon/blob/6c7d84d3e5f08b7115a3a1ab31cdf093bc3fbf5e/src/interval.js#L16

It should check Settings.throwOnInvalid here, and throw if true.

It currently doesn't.

DanielJoyce:
https://codesandbox.io/embed/luxon-playground-50osl  

Playground repro

icambron:
Sure, would take a fix

StheffanyHadlich:
I can work on it

icambron:
Right

icambron:
Fixed in that PR, released in 1.17.0

PR #527, opened 2019-06-19
Add throwOnInvalid exception on interval validation

StheffanyHadlich:
Add throw exception required in #521

### DIFF
--- a/src/interval.js
+++ b/src/interval.js
@@ -87,11 +87,16 @@ export default class Interval {
     const builtStart = friendlyDateTime(start),
       builtEnd = friendlyDateTime(end);
 
-    return new Interval({
-      start: builtStart,
-      end: builtEnd,
-      invalid: validateStartEnd(builtStart, builtEnd)
-    });
+    const validateError = validateStartEnd(builtStart, builtEnd);
+
+    if (validateError == null) {
+      return new Interval({
+        start: builtStart,
+        end: builtEnd
+      });
+    } else {
+      return Interval.invalid(validateError);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/interval.js
```diff
@@ -13,10 +13,14 @@ function validateStartEnd(start, end) {
   } else if (!end || !end.isValid) {
     return new Invalid("missing or invalid end");
   } else if (end < start) {
-    return new Invalid(
-      "end before start",
-      `The end of an interval must be after its start, but you had start=${start.toISO()} and end=${end.toISO()}`
-    );
+    if (Settings.throwOnInvalid) {
```
GillesDebunne: For consistency, the other calls to `new Invalid` in this method should also throw.

Actually, the other `from` constructor of this class (`fromISO`) relies on `Interval.invalid` which does `throw` based on Settings.

`fromDateTimes` should do the same, with something along the lines of:
```
const validationError = validateStartEnd(builtStart, builtEnd)
if (validationError == null) return new Interval({ start: builtStart, end: builtEnd })
else return Interval.invalid(validationError)
```

icambron: Yeah, what Gilles said is right. If you use `Interval.invalid()`, you'll get the `throw` for free there and you won't have to check the settings here.

icambron:
Hi, thanks for the contribution! Can you add a test case and also address the comments?

src/interval.js
StheffanyHadlich: So intead of throwing the exception inside validateStartEnd, I should trhow it when the function is called in case there is an error?

icambron: Yeah, please throw inside of `Interval.invalid()` so that it covers all callers of that, and then call that from here instead of constructing a new instance directly.

StheffanyHadlich: I did this but according to my test I'm receiving an object and not an exception, I'm not sure why

icambron:
I will take a look at why this is failing the tests.

icambron:
Merged in f79c9b5

icambron:
Now in 1.17.0

PR #534, opened 2019-06-29
docs: fix toRelative unit options

Dosant:
This pr fixes misleading docs for unit options in toRelative and toRelativeCalendar methods. 
I will also make a fix for [@types/luxon](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/luxon/index.d.ts#L24) with correct unit options.


icambron:
hmm, yeah, thanks. If you sign the CLA, I'll get this merged.

PR #535, opened 2019-07-01
2.0 fixes and features

### LINKED ISSUE #378: Side-effects of settings defaultZoneName
I came across this whilst trying to mock dates within tests, and found it rather un-intuitive, so thought I'd raise after having a quick check for any mentions previous.

As part of my tests, I wanted to change the timezone, so I changed the value of `Settings.defaultZoneName` in order to accomplish this. To unpatch the setting, I captured its previous value and re-instated, e.g.:

```js
someTestCase(() => {
  let original;

  beforeEach(() => {
    Settings.defaultZoneName = 'Asia/Singapore';
  });

  afterEach(() => {
    Settings.defaultZoneName = original;
  });
}
```

However, after debugging and issue I found that this is not in fact restoring the original state:

* `Settings.defaultZoneName`is [really just a proxy to `Settings.defaultZone`](https://github.com/moment/luxon/blob/master/src/settings.js#L52)
* [`Settings.defaultZone` is initially a `LocalZone` instance]([https://github.com/moment/luxon/blob/master/src/settings.js#L61]) because [the private `defaultZone` variable is set to null](https://github.com/moment/luxon/blob/master/src/settings.js#L7)
* As a result, when you set `Settings.defaultZoneName` to be *anything*, [it is normalized to produce a non-`LocalZone`](https://github.com/moment/luxon/blob/master/src/impl/zoneUtil.js#L13:17)

The end result is that you end up with this, slightly confusing (at least in my mind) behaviour:

```JS
luxon.Settings.defaultZone
// OUT: LocalZone {}

// Set the defaultZoneName to equal its current value
luxon.Settings.defaultZoneName = luxon.Settings.defaultZoneName

luxon.Settings.defaultZone
// OUT: IANAZone {zoneName: "Asia/Singapore", valid: true}
```

My initial thought would be to have the `defaultZoneName` return `null` when the `defaultZone` is a `LocalZone` instance, but I'm not well enough versed in the Luxon codebase to understand the ramifications of that change.

GillesDebunne:
This PR includes the commits I submitted on the 2.0 branch before it was rebased, plus some new ones, including the deprecation of `Settings.defaultZoneName`.

The "Info.hasDST" commit fixes two typos in the method, which would would have easily been caught by typescript. What are your thoughts on moving the repository to TS?

Fixes #378


icambron:
Thanks.

Re:TS: I once took a crack at writing all this in Flow (which I somewhat prefer to TS because it integrates a bit better with Babel) and quickly gave up, because it was taking way too much work to get my internal types working. Luxon uses lots of quick-and-simple internal objects, and I don't want to declare all of them. It's not that I dislike types checking per se (I program regularly in statically typed languages); it's just that my attempts to use them in JS have resulted in huge amounts of boilerplaty stuff. I'm convinceable here, but it has to be really nice and simple.

PR #536, opened 2019-07-03
exception handling added for hasIntl and hasRelative functions in src/impl/util.js

gokulsiva:
hasIntl and hasRelative functions in src/impl/util.js are wrapped with try catch to avoid exceptions in IE11 browsers.


icambron:
Looks good, thanks

icambron:
Released in 1.16.1

Issue #125, opened 2017-12-10
Support [ ] for format literals


icambron:
Luxon currently supports single quotes to mark literals in `toFormat` and `fromString` (meaning, signal to the formatter that these characters shouldn't be consider formatting tokens). But the more I use it, the more I think it's a lot harder to read that Moment's `[ ]` delimeters. So I'd like to support them. (Along with a way of escaping them? Ugh)

icambron:
@benneq in principle I'm for that, needs some careful thought to avoid being a big pain implementation-wise

benneq:
My main (hidden) question was: Why do you prefer square brackets instead of single quotes?

I mean: In a normal app you have 5 to 10 different formats that you store behind some constants and then only use the constants across your app:
```
const DATE_DAY_MONTH = "'Day: 'dd', Month: 'MM";
// or with square brackets:
const DATE_DAY_MONTH = "[Day: ]dd[, Month: ]MM";
// or someone may even prefer this:
const DATE_DAY_MONTH = (luxon) => `Day: ${luxon.day}, Month: ${luxon.month}`;
```
I wouldn't say that one is more readable than the other. It's just different.
You will touch those maybe once or twice a year, not every day. And additionally those format strings are pretty short. Though I'd say that it really doesn't matter what kind of delimiter you use for that. (the same applies to the "optional" delimiter).

As far as formatting goes, I'd say: It's just a matter of taste which style (or delimiter) you prefer. As long as `luxon` provides exactly one way of doing it.

For parsing, it could sometimes be really useful to have optional parts within the token. For example in my current project the API doesn't output the milliseconds if they are `000`. Though I have to parse it using `HH:mm:ss.SSS` and if that fails, try again with `HH:mm:ss`. (Of course it's the same as above with the formatting: There's somewhere a `parseTime` function, which handles this).

icambron:
@benneq they're easier to visually match. I had thought it would be the same but it's really not.

As for optional parsing, we should implement that, though presumably with a different delimiter. Perhaps {}.

icambron:
This is merged into the 2.0 branch, so I'm going to close it

renatoargh:
My reason for preferring brackets to quotes is that because by using them you don't have to escape like this:

`'dd/MM/yyyy \'at\' HH:mm \'(\'ZZZZ\')\''`

adamreisnz:
> My reason for preferring brackets to quotes is that because by using them you don't have to escape like this:
> 
[...]

PR #544, opened 2019-07-17
#125 Literal parsing/formatting using brackets

GillesDebunne:
This PR replaces simple quotes with brackets to create string literals #125 

I find brackets easier to read for a human, and they are already used in moment for this purpose.

Nested brackets are possible, if they are balanced (`'HH[[:]]mm' -> 23[:]45`).

However, escaping brackets, as in `\[` or `\]` was not implemented.
1. The `\` in `'\['` is already caught by JS, so this is equivalent to a single `'['`. One would need to use `'\\['`which is confusing.
2. The `parseFormat` automata would become much more complex.


src/datetime.js
```diff
@@ -1081,7 +1081,7 @@ export default class DateTime {
    */
   get isInDST() {
     return this.c
-      ? this.offset > this.set({ month: 1 }).offset || this.offset > this.set({ month: 5 }).offset
+      ? this.offset > this.set({ month: 12 }).offset || this.offset > this.set({ month: 6 }).offset
```
icambron: Found a bug here?

icambron:
This looks great!

src/datetime.js
GillesDebunne: Nope. That should have been a different PR.

This is to make this code similar to what is done in `Info.hasDST()`

PR #547, opened 2019-07-28
Add ability to parse some of localized macro format tokens

### LINKED ISSUE #115: Supporting macro tokens in the parser
`DateTime#toFormat` supports "macro tokens" like `DD` or `TTT`. See the bottom of the table [here](https://moment.github.io/luxon/docs/manual/formatting.html#table-of-tokens) to see them. In principle, these could be supported by the parser too, but it's a bunch of work and I don't feel like doing it. I'm also not willing to let the library grow much bigger and more complex unless there's a lot of interest in the feature. Anyway, this is an outline of how I think it *could* work.

The goal is to take a format like `D` and "expand" it to something like `MM/dd/yy` or `dd/MM/yy` based on the locale. The steps below work though that process, though I'm not saying each intermediate object must really be generated.

Step 1: Look up the token in the `Formats` class and pull out the Intl opts the correspond to it. So we get something like

```js
{
  year: 'numeric',
  month: 'numeric',
  day: 'numeric'
}
```

Step 2: use `formatToParts` with a dummy date to generate a list of tokens using those opts. 

```js
new Intl.DateTimeFormat('fr', fmt).formatToParts()
[ { type: 'day', value: '06' },
  { type: 'literal', value: '/' },
  { type: 'month', value: '12' },
  { type: 'literal', value: '/' },
  { type: 'year', value: '2017' } ]
```

Step 3:  We can map that back the input to get a list of almost-tokens. In the example below, note the new style key, which I've injected from looking up each unit in the original object. Also note that I've removed `value`; we don't care about that, just the ordering.

```js
[ { type: 'day', style: 'numeric'},
  { type: 'literal', value: '/'},
  { type: 'month', style: 'numeric'},
  { type: 'literal', value: '/' },
  { type: 'year', style: 'numeric' } ]
```

Step 4: Next, we need to know the tokens these correspond to. The inverse of this mapping is in the `TokenFormatter` class already as `unitForToken`, but we need `tokenForUnit` that takes in the unit ('day', etc) and the style ('numeric', 'short', etc) and returns a token like 'MM'. Now you know what token goes which each piece of the format. And you can glue that all together:

```js
'dd/MM/yy'
```

Step 5: This all needs wrapped into a function for expanding the string before the main parsing steps in, with a name like `expandMacroTokens`. It should return the string as-is if there are no macro tokens in it. It should process the string one token at a time; there might be other tokens, other literals, and even multiple macros in it. It should be called by the parsing code in a way that `fromStringExplain()` can return it explicitly as one of the fields.

futpib:
Fixes #115

### DIFF
--- a/src/impl/formatter.js
+++ b/src/impl/formatter.js
@@ -14,7 +14,7 @@ function stringifyTokens(splits, tokenToString) {
   return s;
 }
 
-const tokenToObject = {
+const macroTokenToFormatOpts = {
   D: Formats.DATE_SHORT,
   DD: Formats.DATE_MED,
   DDD: Formats.DATE_FULL,
@@ -78,6 +78,10 @@ export default class Formatter {
     }
 
     return splits;
+  }
+
+  static macroTokenToFormatOpts(token) {
+    return macroTokenToFormatOpts[token];
   }
 
   constructor(locale, formatOpts) {
@@ -150,9 +154,9 @@ export default class Formatter {
               "weekday"
             ),
       maybeMacro = token => {
-        const macro = tokenToObject[token];
-        if (macro) {
-          return this.formatWithSystemDefault(dt, macro);
+        const formatOpts = Formatter.macroTokenToFormatOpts(token);
+        if (formatOpts) {
+          return this.formatWithSystemDefault(dt, formatOpts);
         } else {
           return token;
         }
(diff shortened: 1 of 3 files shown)


icambron:
This looks good. I'm going to pull it down locally and give it a spin. I'll add comments when I get a chance to do that.

Thanks for the contribution!

icambron:
@futpib Sorry it took me *so* long to get to this, but I played around with it and it's great. Released in 1.18.0! Thanks again.

PR #564, opened 2019-08-29
fix inconsistency within constructor

NathanCanine:
const x = require("luxon").DateTime;

console.log(x.local().toISO());
console.log(x.local(undefined).toISO());
console.log(x.local(null).toISO());

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -440,9 +440,12 @@ export default class DateTime {
    * @return {DateTime}
    */
   static local(year, month, day, hour, minute, second, millisecond) {
-    if (isUndefined(year)) {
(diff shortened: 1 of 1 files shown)


icambron:
No, this isn't an inconsistency. The way JavaScript is structured, you expect `foo(undefined)` to work the same way as `foo()`. Callers often take advantage of this, as in `foo(someBool ? someArg : undefined)`. That's why JS passes in undefined to the function in the first place.

In other words, this change swims against normal expectations in JavaScript and reduces the caller's flexibility. Perhaps this is a flaw in JS itslef, but Luxon should not be what fixes it.

PR #567, opened 2019-09-03
Mark first parameters in DateTime.local() and DateTime.utc() as optional

### LINKED ISSUE #566: WebStorm shows false warning on DateTime.local()
When using WebStorm (and presumably other JetBrains IDEs), the built-in jsdoc parser shows a false warning when using `DateTime.local()` with no arguments:

image

I'm working on a PR now to mark the `year` parameter as optional, since according to the docs that is a valid use.

tdukart:
Since the `year` parameters are officially optional, this hints to IDEs that it's safe to use empty calls like `DateTime.local()`.

Fixes #566.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -422,7 +422,7 @@ export default class DateTime {
 
   /**
    * Create a local DateTime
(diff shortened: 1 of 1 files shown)


icambron:
Thanks!

PR #572, opened 2019-09-13
Handle undefined DTF locale

### LINKED ISSUE #571: Errors on FB App Windows 10
Hard for me to know about support, but we are consistently getting windows10 issues with this report:

https://sentry.io/share/issue/feea7007054f41b5954623ba1a09caf8/

Let me know if you need more info to reproduce, on our end unfortunately we don't have enough time to debug it, sorry :(

GillesDebunne:
Trying to fix #571, no easy repro case. Specs seem to imply that locale should always be defined though, but since there was already a special case for 'und'...


icambron:
This is a perfectly reasonable change

icambron:
Released in 1.17.3

Issue #563, opened 2019-08-28
Bunch of NaNs instead of invalid DateTime


balagge:
Adding a long period to a DateTime object may produce invalid DateTimes, because of "overflow".
However, sometimes the `invalid` flag is not set; nevertheless, the DateTime object is unusable (contains `NaN`s)

The following example calculates the last valid DateTime objects on the timeline, and the "first" invalid ones. For the allowed range of Dates in javascript, see https://www.ecma-international.org/ecma-262/6.0/#sec-time-values-and-time-range, which says:

> –100,000,000 days to 100,000,000 days measured relative to midnight at the beginning of 01 January, 1970 UTC.

```javascript
const orig = DateTime.utc(1970,1,1,0,0,0,0); 
const N = 1E8 // max range of Dates in days
const secs = 24*60*60; // secs per day

// last valid DateTimes
const a = orig.plus({days: N}) // should be valid (it is)
console.log("a: ", a)

const b = orig.plus({seconds:N * secs}) // should be valid (it is)
console.log("b: ", b)


// first invalids
const c = orig.plus({days: N+1}) // should be invalid (it is)
console.log("c: ", c)

const d = orig.plus({seconds:N * secs+1}) // should be invalid (it is not, but a corrupt object)
console.log("d: ", d)

const e = orig.plus({seconds:(N +1)* secs}) // should also be invalid (it is not, but a corrupt object)
console.log("e: ", e)
```

**Expected behavior:** `a` and `b` should be valid, `c`, `d`, `e` should be invalid.

**Actual behavior:** 
 - `a` and `b` are valid, `c` is invalid (as expected)
 -  `d` and `e` are reported as valid, but are actually corrupt objects (error)

neversaid:
I have added a PR for this issue. Details are written down on this as well #576

icambron:
Ah, I hadn't read the OP carefully enough. See also #299

icambron:
Fixed by #576

PR #576, opened 2019-09-22
Fix for invalid Datetime when not in valid ECMA DateTime range

neversaid:
As reported in #563 this problem occurs when with the DateTime.plus a duration of seconds is added which will result in a DateTime Object outside of the ECMA DateTime range.

Reproduction Code 
```js
const orgDate = DateTime.utc(1970,1,1,0,0,0,0); 
const maxRangeDays = 1E8 // max range of Dates in days
const secsPerDay = 24*60*60; // secs per day

// When adding Seconds
// d.invalid is null - expected to have reason "invalid input"
// d.c contains NaN - expected to be null
const d = orgDate.plus({seconds: maxRangeDays  * secsPerDay +1})

// When adding days
// e.invalid is has reason "invalid input"
// e.c contains null
const e = orgDate.plus({day: maxRangeDays  +1})
```

The reason why the seconds did not result within an invalid result is because:

datetime.js -> adjustTime 

In this function two different ways are used to create the returning ts property.
Duration Properties of type "hours, minutes, seconds, milliseconds" are parsed by Durations.fromObject -> as("milliseconds")
whereas Duration Properties of type "year,month,day" are parsed by the objToLocalTS.

In the case we have only  "year,month,day" the function objToLocalTS already returns an invalid object with NaN for the properties, whereas the "timepart" durations are passed as an number to the constructor of DateTime.

Because there is no "test" if the Object created in the constructor by function tsToObj is valid, the new DateTime returned appears valid (but it isnt because its full of NaN and the Date,UTC in the tsToObj already returned invalid date!).

Because of this, there was an additional check in the constructor added of the DateTime class.
This invalidates the c and o, if the returned Object from tsToObj is not valid anymore.

For further refactoring task i would take the discussion for the functions:
DateTime.tsToObj
DateTime.objToLocalTS

IMHO this functions should throw an error which can then be catched if Date.XXX is not giving back a successfull result.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -370,11 +370,12 @@ export default class DateTime {
    * @access private
    */
   constructor(config) {
-    const zone = config.zone || Settings.defaultZone,
-      invalid =
-        config.invalid ||
-        (Number.isNaN(config.ts) ? new Invalid("invalid input") : null) ||
-        (!zone.isValid ? unsupportedZone(zone) : null);
+    const zone = config.zone || Settings.defaultZone;
+
+    let invalid =
+      config.invalid ||
+      (Number.isNaN(config.ts) ? new Invalid("invalid input") : null) ||
+      (!zone.isValid ? unsupportedZone(zone) : null);
     /**
      * @access private
      */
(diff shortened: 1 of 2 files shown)


icambron:
This makes sense to me. Thanks!

PR #577, opened 2019-09-22
[feature] caching of offset in DateTime moment#506

neversaid:
This will avoid fetching unnecessary often from zone.offset - we already set the offset in the constructor to the private property o on DateTime.


neversaid:
Thx!
Jepp, honestly i am not that happy with the current constructor in regards of readability :).
But i already had a peak to the v2 - if you want me to help for v2 (doing most of the time typescript ;) ) let me know.

src/datetime.js
```diff
@@ -1114,7 +1114,7 @@ export default class DateTime {
    * @type {number}
    */
   get offset() {
-    return this.isValid ? this.zone.offset(this.ts) : NaN;
```
icambron: Why does this need a `+`? Seems like it would already be a number?

neversaid: but i can be null as well if in some crazy case the invalid is false...

GillesDebunne:
> But i already had a peak to the v2 - if you want me to help for v2 (doing most of the time typescript ;) ) let me know.

Sure, feel free to jump in, any comment on the code is welcome!

icambron:
Thanks for the contribution!

PR #581, opened 2019-09-27
Format override example has incorrect code

rxgx:
The example code for `Intl.DateTimeFormat` uses the incorrect order of params for `Object.assign`


icambron:
Ha, yes. Thanks

PR #589, opened 2019-10-05
Add method Duration#times

### LINKED ISSUE #557: Duration multiply method
In our app, we ran into a case where we need to multiply a Luxon `Duration` object by a number before adding it to another number. It would be awesome if there were a multiply method that took in a number and returned the duration multiplied by that number. Here's a simplified example of what we'd like:

``` javascript
Duration.fromObject({ hour: 1 }).multiply(5);
```

I realize in this example you could put the 5 in the object, but in our case, it makes sense to have this behavior.

benblank:
This PR introduces a new method, `Duration#times()`, which multiplies a Duration by a scalar.

The original feature request (#557) called the proposed method `multiply`, but I felt `times` was more consistent with the existing `plus` and `minus` methods.

Closes #557.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -478,6 +478,26 @@ export default class Duration {
   }
 
   /**
+   * Scale this Duration by the specified amount. Return a newly-constructed Duration.
+   * @param {number} scalar - The amount to multiply by.
+   * @example Duration.fromObject({ hours: 1, minutes: 30 }).times(2) //=> { hours: 2, minutes: 60 }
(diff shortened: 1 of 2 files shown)


icambron:
Thanks for the contribution. I like `times` too. However, after some consideration, I'm closing this in favor of #592 

You can implement this easily with that, something like:

```js
Duration.prototype.times = function(scalar) { return this.mapUnits(x => x * scalar); }
```

PR #590, opened 2019-10-05
add duration multiply method

jdim:
Hello, it is my first contribution.
This PR solves the problem in issue #557
Just add a new method and some tests.


icambron:
Thanks for the contribution. After some consideration, I'm closing this in favor of #592, which I based on this.

Issue #411, opened 2019-01-21
Documentation rendering bug


benmccann:
On https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html the description of `fromMillis` is truncated the first place it appears on the page. If you click that link to take you further down then you can see the whole description

icambron:
Yes, this is a bug in ESDoc, as far as I can tell.

WretchedDade:
Hi @icambron, looks like the the reason that message and other messages get truncated is because of the 'i.e.'. When playing around with the text I found that the summary messages gets truncated after the second period. We could resolve this by removing the use of 'i.e.', thoughts?

icambron:
Yeah, that seems reasonable. Could be changed from "i.e." to "meaning"

WretchedDade:
@icambron, I opened a PR where I replaced i.e. with meaning. I went ahead and made the change in overall rather than just the summary of fromMillis.

PR #597, opened 2019-10-20
Fixed truncated summary issue

WretchedDade:
ESDoc truncates summary messages after two periods, therefore the use of 'i.e.' would cause summaries to get unusually truncated. To fix this, I ran a find and replace, replacing 'i.e.' with meaning as suggested by @icambron.

This PR is in reference to #411


icambron:
Thanks!

Issue #595, opened 2019-10-15
fromFormat and fromFromatExplain does not support Quarter values


el:
```
DateTime.fromFormatExplain("2017Q1","yyyyQq")
```
and
```
DateTime.fromFormatExplain("2017Q01","yyyyQqq")
```
should create a date without a problem. But they don't.

icambron:
Quarters aren't currently supported in the parser, which isn't completely symmetrical with the formatter. You can see the table of tokens here: https://moment.github.io/luxon/docs/manual/parsing.html#table-of-tokens

I think this is a reasonable feature request, though. I'd take a pull request adding quarter support.

ibarrae:
Hey folks @el @icambron , given the examples above, what should be the expected outcome?

Maybe it's something like this?
`DateTime.fromFormatExplain("2017Q1", "yyyyQq") === 2017-01-01T00:00:00.00:TZ`

icambron:
You want to change `fromFormat()`; `fromFormatExplain` is a debugging tool. Also, it will parse to a local time by default. So you want something like:

```
DateTime.fromFormat("2017Q1", "yyyy'Q'q", { zone: "utc" }).toISO() //=> "2017-01-01T00:00.00Z"
```

el:
I think `luxon` should have a configuration for quarters. It is common practice to have a different definition for quarters. What you are suggesting is called [Calendar Year](https://en.wikipedia.org/wiki/Calendar_year), [Fiscal Year](https://en.wikipedia.org/wiki/Fiscal_year) has a different start/end dates.

icambron:
Luxon will not support customizable fiscal calendars, sorry.

icambron:
Fixed by #602

PR #602, opened 2019-10-25
Add quarter parsing support

el:
This PR probably has missing things, but it works so this should be a good start 
This is an attempt to support #595

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -119,6 +119,10 @@ function unitForToken(token, loc) {
           return intUnit(two);
         case "m":
           return intUnit(oneOrTwo);
+        case "q":
+          return intUnit(oneOrTwo);
+        case "qq":
+          return intUnit(two);
         case "s":
           return intUnit(oneOrTwo);
         case "ss":
@@ -294,6 +298,8 @@ function dateTimeFromMatches(matches) {
         return "weekNumber";
       case "k":
         return "weekYear";
+      case "q":
+        return "quarter";
       default:
         return null;
     }
@@ -306,6 +312,10 @@ function dateTimeFromMatches(matches) {
     zone = IANAZone.create(matches.z);
   } else {
     zone = null;
+  }
+
+  if (!isUndefined(matches.q)) {
+    matches.M = (matches.q - 1) * 3 + 1;
   }
 
   if (!isUndefined(matches.h)) {
--- a/test/datetime/tokenParse.test.js
+++ b/test/datetime/tokenParse.test.js
@@ -217,6 +217,21 @@ test("DateTime.fromFormat() parses format month names", () => {
   expect(i.day).toBe(25);
 });
 
+test("DateTime.fromFormat() parses quarters", () => {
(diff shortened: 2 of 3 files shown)


test/datetime/tokenParse.test.js
```diff
@@ -217,6 +217,21 @@ test("DateTime.fromFormat() parses format month names", () => {
...
+  expect(i.month).toBe(4);
+  expect(i.quarter).toBe(2);
+  expect(DateTime.fromFormat("2019Q1", "yyyy'Q'q").month).toBe(1);
+  expect(DateTime.fromFormat("2019Q2", "yyyy'Q'q").month).toBe(4);
+  expect(DateTime.fromFormat("2019Q3", "yyyy'Q'q").month).toBe(7);
+  expect(DateTime.fromFormat("2019Q4", "yyyy'Q'q").month).toBe(10);
+  expect(DateTime.fromFormat("2019Q01", "yyyy'Q'qq").month).toBe(1);
+  expect(DateTime.fromFormat("2019Q02", "yyyy'Q'qq").month).toBe(4);
+  expect(DateTime.fromFormat("2019Q03", "yyyy'Q'qq").month).toBe(7);
+  expect(DateTime.fromFormat("2019Q04", "yyyy'Q'qq").month).toBe(10);
```
mrcrow85: @el It would be nice to add tests for wrong dates like `"2019Q05"` and `"2019Q5"`

icambron: I agree

src/impl/tokenParser.js
```diff
@@ -308,6 +314,10 @@ function dateTimeFromMatches(matches) {
     zone = null;
   }
 
+  if (!isUndefined(matches.q)) {
+    matches.M = (matches.q - 1) * 3 + 1;
```
icambron: The unfortunate part of this is that the error message for specifying a quarter that isn't between 1-4 is that it will complain that the _months_ are out of range. I think we can live with that though. The alternative would be to make quarters a first-class unit in Luxon, which seems painful.

icambron:
This looks good. If you add some additional tests showing that invalid quarter strings yield invalid DateTime instances, I'll get this merged.

test/datetime/tokenParse.test.js
el: Done 👍

src/impl/tokenParser.js
el: I actually tried this:
```js
  if (!isUndefined(matches.q)) {
    if (matches.q < 1 || matches.q > 4)
    {
        throw new Error(`quarter value "${matches.q}" is out of 1-4 range`);
    }
    matches.M = (matches.q - 1) * 3 + 1;
  }
```
but thrown errors are not caught.

icambron: Yeah, it's not the place where the validation happens; it happens later once the object has been assembled. That's the weird thing about Q just translating to M; the Q context gets lost by the time the validation step runs. But it is what is.

icambron:
Thanks!

PR #605, opened 2019-10-28
Fix Duration#toISO floating point madness

### LINKED ISSUE #596: Duration toISO produces invalid ISO string
Hey 👋 

There's some floating point madness going on that makes Duration produce invalid ISO strings. See example below.

```js
const duration = Duration.fromObject({seconds: 17, milliseconds: 548});
const isoString = duration.toISO() // -> "PT2M17.548000000000002S"
const parsedDuration = Duration.fromISO(isoString);
console.log(parsedDuration.isValid); // -> false
```

mrcrow85:
This will fix #596 by handling [floating point madness](https://stackoverflow.com/questions/588004/is-floating-point-math-broken) when working with seconds and milliseconds.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -1,10 +1,10 @@
-import { isUndefined, isNumber, normalizeObject, hasOwnProperty, asNumber } from "./impl/util.js";
-import Locale from "./impl/locale.js";
-import Formatter from "./impl/formatter.js";
-import { parseISODuration } from "./impl/regexParser.js";
-import Settings from "./settings.js";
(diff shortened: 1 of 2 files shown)


icambron:
Take a look at `Utils.roundTo`. Looks like you could use that?

mrcrow85:
> Take a look at `Utils.roundTo`. Looks like you could use that?
> 
> https://github.com/moment/luxon/blob/969246ac657712f198f08421c8e683f2cf62ea98/src/impl/util.js#L127

@icambron Done. Thanks for pointing that out :slightly_smiling_face:

icambron:
Thanks!

PR #612, opened 2019-11-03
2.0 take 2 merge

GillesDebunne:


icambron:
I just skimmed it, but this looks good to me.

GillesDebunne:
Did the merge affect the code coverage, or was this 2.0 branch already behind?

i.e. shall we merge it as is, or did the merge break something and we need to fix the codecov first?

icambron:
@GillesDebunne I think it is saying that "91.66%" of code that changed has coverage, whereas 97.84% of the whole project now has coverage. I don't think it's cause for concern; it's just that a lot of the untested code changed.

icambron:
Oh, I know why your test is failing in CI. Let me see if I can fix that

icambron:
OK, I took a look. That test is failing because Node 10 doesn't actually support RelativeTimeFormat, so it just uses the English fallback, which is also why I didn't write the test in the first place. It passes in Node 12, but three other tests fail because of Intl changes.

I'd prefer to get 2.0 out the door and then do the Node upgrade, just so we have less confusion to merge.

GillesDebunne:
Yes, that's what I thought. That was a test to improve code coverage.

I'll just revert my changes in locale.js, it was a simple typo fix. That
way we can merge this, and I can re-start working on the real task, which
is to rebase the typescript branch on top of it. I have little time right
now, but I have this as a background task.

PR #630, opened 2019-12-18
Remove Meteor from install docs

qualitymanifest:
Meteor has had support for npm for a few years now, and package installation and use is the same as for Webpack. Just replaced Moment with Luxon in my Meteor app, very easy transition and the reduction in bundle size is great.


icambron:
Thanks!

Issue #617, opened 2019-11-11
Throw an error if 24-hour hour and meridiem are both specified


icambron:
If a user does this:

```js
DateTime.fromFormat('930PM', 'Hmma').toISOTime() //=> '09:30:00.000-08:00'
```
...they almost certainly meant `h`, not `H` and it isn't easy to debug (see #616). When both are specified, we should just throw a ConflictingSpecificationError, like we do if you specify the wrong day of the week.

PR #633, opened 2019-12-23
Throw error when specifying 24-hour format with meridiem

qualitymanifest:
Closes https://github.com/moment/luxon/issues/617

I'm getting two failing tests but I think they are unrelated.

<details>
  <summary>Failing tests</summary>
  
Selection_252

</details>

I've tested it out in the browser with the line of code from that issue and it throws as it should


qualitymanifest:
Hmm, would it be better to put it [here](https://github.com/moment/luxon/blob/5bf5f2b04a0d16417c35e2c1f71348d559a4de2b/src/datetime.js#L802) with `if (fmt.includes("a") && fmt.includes("H")) { ...` ?

icambron:
No, I think it's better where you put it in the pull. Going to merge this.

Thanks!

Issue #394, opened 2018-12-31
`toRelativeCalendar` doesn't really work like you'd expect


icambron:
This is a little funny:

```js
base = DateTime.local(2018, 12, 31);
base.plus({ days: 1 }).toRelativeCalendar() //=> "next year"
```

This is technically correct, but I suspect most people would want that to say "tomorrow". Same issue with the end of months.

The question is: when *should* it say next year?

 1. whenever it is next year? (current behavior)
 1. never?
 1. When it's over 1 year in the future? (this is pretty weird IMO; in our 12/31 example, it would only say "next year" for one day at the end of next year")
 1. when it's over N months in the future? For what N?

Same problem for months. And I haven't even introduced weeks, which is surely similar.

djhvscf:
Optional parameter? Let's say: we define a default value 'next year' but the developer can send a parameter which indicates the value to be return

robrich:
I'd say the closest boundary is the correct one, but overriding it with a parameter sounds interesting.  I grant it's likely a major version bump, but I'd say if it's within 7 days and hits such a boundary, it should say "tomorrow, if it's within 8 days and hits a boundary, "next week", if it's within 32 days and hits a boundary, "next month".  It sounds like we need to decide what "correct" is before we PR, right?

icambron:
Yes, we need a spec. I like the idea that the threshold matches the size of the unit. I don't like the idea of overrides because I think it would be a confusing API. We can always add it later but we can't take it back once we do. So I'd like to make this change without expanding the API surface area.

I don't think this really requires a major version bump, either.

FagnerMartinsBrack:
Here's a suggestion that I'm using in one of my projects where I face this problem frequently. I run some TDD in it that's why you can see some pattern generalization.

Feel free to use it as much as MIT and WTFPL allows you :P

**TL;DR:**

If it's one or more years apart, say next year or x years
If it's one or more months apart, say next month or x months
if it's one of more days apart, say tomorrow or x days

Anything that is less than 1 day apart is counted as "in x hours", "in x minutes"... and so forth

I would appreciate if someone could help to fix the `void(0)` loose branch

```javascript
const toRelativeYears = years => {
  return years === 1
    ? 'next year'
    : `in ${years} years`;
};

const toRelativeMonths = months => {
  return months === 1
    ? 'next month'
    : `in ${months} months`;
};

const toRelativeDays = days => {
  return days === 1
    ? 'tomorrow'
    : `in ${days} days`;
};

[...]

PR #637, opened 2019-12-31
.html -> .md

robrich:
On the website it's correctly linking to install.html.  On GitHub's markdown rendering, it yields a 404.

### DIFF
--- a/docs/intl.md
+++ b/docs/intl.md
@@ -10,7 +10,7 @@ DateTime.local()
 
 ## Making sure you have access to other locales
 
-Please see the [install guide](install.html) for instructions on making sure your platform has access to the Intl APIs and the ICU data to power it. This especially important for Node, which doesn't ship with ICU data by default.
+Please see the [install guide](install.md) for instructions on making sure your platform has access to the Intl APIs and the ICU data to power it. This especially important for Node, which doesn't ship with ICU data by default.
 
 ## How locales work
 


icambron:
Won't this break the website, though? In general, the markdown doesn't "work"; it only works when compiled. For example, the links into the docs don't work either.

I tried to capture that here: https://github.com/moment/luxon/blob/master/docs/readme.md

I'd of course love to fix it so that it works in both places.

robrich:
Good point.  And yes, making this change here will break the "compiled" versions.  Perhaps the compilation step could rewrite `.md` to `.html` on the way through?  How do I run the compilation locally to understand the feasability of this?

robrich:
Also, the build fail is really interesting.  If running from `Dec 31` and adding 1 day, the relative date says "next year" not "tomorrow".  Is this issue-worthy?

icambron:
`npm run docs`. In theory you can do it with an esdoc plugin. Last time I tried to do that, I ran into trouble, but I don't recall why

icambron:
The issue about "next year" is #394

robrich:
Not finding the feature in esdoc docs, I asked https://github.com/esdoc/esdoc/issues/558

icambron:
In my experience, you are unlikely to get a response from the esdoc maintainer.

icambron:
I wonder if for 2.0, we should scrap esdoc and use typedoc or similar. Then the conversation about how to deploy the markdown gets reset anyway

GillesDebunne:
I have not tried typedoc or other alternatives, but a typescript-aware doc
would be nice.

icambron:
Going to close this one for now. Let's revisit when we do 2.0.

PR #646, opened 2020-01-09
Remove `true` param from Interval.length example

mrcoles:
Looking at the source code for Interval.length there is no second parameter: https://github.com/moment/luxon/blob/master/src/interval.js#L208

```
  length(unit = "milliseconds") {
    return this.isValid ? this.toDuration(...[unit]).get(unit) : NaN;
  }
```

In this page of the docs it has `i.length('years', true)`. Maybe this is an old API? I removed it from the example here.


icambron:
Thanks!

icambron:
Yup, thanks!

PR #660, opened 2020-02-08
Add support for ISO basic format to DateTime#toISO

### LINKED ISSUE #656: Add support for ISO 8601 basic format to DateTime.toISO()
From [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601):

> Representations can be done in one of two formats – a basic format with a minimal number of separators or an extended format with separators added to enhance human readability. The standard notes that "The basic format should be avoided in plain text." The separator used between date values (year, month, week, and day) is the hyphen, while the colon is used as the separator between time values (hours, minutes, and seconds). For example, the 6th day of the 1st month of the year 2009 may be written as "2009-01-06" in the extended format or simply as "20090106" in the basic format without ambiguity.

This could be done as a `{ format: 'basic' }` option with the default being `{ format: 'extended' }`. This format is useful when used in a URL.

mohd-akram:
Closes #656.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -179,10 +179,10 @@ function parseDataToDateTime(parsed, parsedZone, opts, format, text) {
 
 // if you want to output a technical format (e.g. RFC 2822), this helper
 // helps handle the details
-function toTechFormat(dt, format) {
+function toTechFormat(dt, format, allowZ = true) {
   return dt.isValid
     ? Formatter.create(Locale.create("en-US"), {
-        allowZ: true,
+        allowZ,
         forceSimple: true
       }).formatDateTimeFromString(dt, format)
     : null;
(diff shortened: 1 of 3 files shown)


mohd-akram:
The examples under `DateTime#toISOTime` don't work.

icambron:
This looks good. I'm not sure what's up with TravisCI. I'll take a look

icambron:
If you rebase onto master, the build should work now.

mohd-akram:
Done.

icambron:
> The examples under DateTime#toISOTime don't work

thanks, fixed in `master`

icambron:
That's for the contribution! I'll release this soon

mohd-akram:
Thanks!

Issue #678, opened 2020-03-22
Error message suggestion


benmccann:
I got the error below:
```
Uncaught Error: fromMillis requires a numerical input
    at Function.fromMillis (luxon.js:6399)
```

I would have expected my code was passing a numerical input, so I'm curious what it's passing instead. It might be nice if the error message contained the given input. E.g. if it said something like `fromMillis requires a numerical input, but received foobar`

icambron:
Sure, makes sense. I'd accept a PR spelling that out

PR #680, opened 2020-03-23
Add more details to exception message

benmccann:
Closes https://github.com/moment/luxon/issues/678

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -555,7 +555,9 @@ export default class DateTime {
    */
   static fromMillis(milliseconds, options = {}) {
     if (!isNumber(milliseconds)) {
-      throw new InvalidArgumentError("fromMillis requires a numerical input");
+      throw new InvalidArgumentError(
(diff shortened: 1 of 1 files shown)


src/datetime.js
```diff
@@ -555,7 +555,9 @@ export default class DateTime {
    */
   static fromMillis(milliseconds, options = {}) {
     if (!isNumber(milliseconds)) {
-      throw new InvalidArgumentError("fromMillis requires a numerical input");
+      throw new InvalidArgumentError(
+        "fromMillis requires a numerical input, but received " + milliseconds
```
icambron: If `milliseconds` is a string, like "123", this will be a confusing message. Perhaps more like

```
throw new InvalidArgumentError(
  `fromMillis requires a numerical input, but received a ${typeof milliseconds} with value ${milliseconds}`)
```

icambron:
thanks!

Issue #681, opened 2020-03-26
Duration.fromISO cant handle leading minus sign


backbone87:
`Duration.fromISO('-P1D')` -> invalid duration
`Duration.fromISO('P-1D')` -> works

a similar issue was resolved in moment:
https://github.com/moment/moment/issues/2408
-> if the string starts with a minus sign, then all component signs are flipped (so `-P1D = P-1D`, `P1D = -P-1D`)

PR #683, opened 2020-04-02
Duration from ISO leading minus

jmadelaine:
This PR fixes [moment/luxon#681](https://github.com/moment/luxon/issues/681).

The issue above links to a similar issue with moment ([moment/moment#2408](https://github.com/moment/moment/issues/2408)) which was resolved.

Although the [ISO 8601 duration format](https://www.digi.com/resources/documentation/digidocs/90001437-13/reference/r_iso_8601_duration_format.htm) does not mention support for a leading minus, there are many other libraries that support ISO duration parsing in this way.

The assumed rule is that a leading minus sign inverts the sign of all values.

Examples of leading-minus ISO strings and their equivalent non-leading-minus ISO strings:

 -P1D == P-1D
  -P5Y3M == P-5Y-3M
  -P-5Y-3M == P5Y3M
  -P-1W1DT13H-23M34S == P1W-1DT-13H23M-34S

### DIFF
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -130,14 +130,6 @@ export function roundTo(number, digits, towardZero = false) {
   return rounder(number * factor) / factor;
 }
 
-export function invertNumber(num) {
-  if (!isNumber(num) || Number.isNaN(num)) {
-    return undefined;
-  } else {
-    return num * -1;
-  }
-}
-
 // DATE BASICS
 
 export function isLeapYear(year) {
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -130,6 +130,14 @@ export function roundTo(number, digits, towardZero = false) {
   return rounder(number * factor) / factor;
 }
 
+export function invertNumber(num) {
+  if (!isNumber(num) || Number.isNaN(num)) {
+    return undefined;
+  } else {
+    return num * -1;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "luxon",
-  "version": "1.22.2",
+  "version": "1.23.0",
```
icambron, line 3: Please remove this, I do that separately at release time.

src/impl/util.js
```diff
@@ -130,6 +130,14 @@ export function roundTo(number, digits, towardZero = false) {
   return rounder(number * factor) / factor;
 }
 
+export function invertNumber(num) {
```
icambron, line 133: I don't think this is needed. `-num` works as expected and NaN passes through it cleanly. And the results of `parseInteger` are either a number or NaN anyway. I propose just doing this directly in regexParser.js

```js
const maybeNegate = num => hasNegativePrefix ? -num : num;
//...
years: maybeNegate(parseInteger(yearStr))
// etc
```

package.json
jmadelaine: I've removed the version bump.

icambron:
Looks good, thanks! I'll get this released very soon.

icambron:
Releases as 1.23.0

PR #704, opened 2020-05-04
Remove console.log

Matrix159:
This console.log is causing a lot of console spam while using luxon.


icambron:
@Matrix159 I just released 1.24.1 with this fix.

PR #716, opened 2020-05-26
Add DATE_MED_WITH_WEEKDAY preset

saltire:
Hi! I recently had a need for a date string with weekday, so I've added one. I noticed that a DATETIME_MED_WITH_WEEKDAY preset got added a while ago, so I followed that example.

Let me know if anything is missing!


GillesDebunne:
Thanks for this contribution, this looks good to me.

fwiwDev:
This is good but not in v2 docs here where the "full set" of presets are listed.

https://github.com/moment/luxon/blob/master/docs/formatting.md#presets

[PR 1029](https://github.com/moment/luxon/pull/1029) was added to fix this.

PR #718, opened 2020-05-31
All tests assume America/NewYork timezone

GillesDebunne:
Some tests require the local timezone to be "America/New_York", relying on the DST of this zone, such as :

_DateTime#diff handles fractional days as fractions of those specific days_

while 2 tests are protected by `if (DateTime.local().zoneName === "America/New_York")`

It seems more consistent to assume all tests require this specific timezone, as is done in `test/zones/local.test.js`:

```
expect(LocalZone.instance.name).toBe("America/New_York");
```


icambron:
Yeah, they do. I do have the `./scripts/test` script set up to set that. But I think this change makes sense anyway.

PR #720, opened 2020-05-31
Add missing values in conversion matrices

GillesDebunne:
And a test to check all values are defined.

Note the conversion `toObject` in the test :

`Duration.fromObject({ seconds: Nan }).get('seconds')` returns 0, which may be considered a bug.


icambron:
Ha! I'd never noticed that.

PR #721, opened 2020-05-31
Remove normalization in ShiftTo

GillesDebunne:
This is a serious change, but I assume the unit tests around `shiftTo` validate the non regression.


icambron:
Yeah, I've been wanting to make this change for a while but kept wimping out. Let's just do it.

Issue #669, opened 2020-02-24
dateTime.plus({hours:24}) works unexpectedly when dateTime is 23 hours prior to dst switch


kaftand:
This error comes up seemingly only when a datetime is 23 hours prior to dst switch. The following example using local time is based in "America/Chicago"
 `var dateTime= lux.DateTime.local(2019,3,9,3,0,0,0);
dateTime.plus({hours:24})`
returns 2019-03-10T03:00, but I would expect 2019-03-10T04:00, since an hour is skipped due to the dst switch. This is highlighted by the fact that  `dateTime.plus({hours:23})` returns the same datetime as `dateTime.plus({hours:24})` for this case. I was able to revert to release 1.17.3, and plus worked as I would have expected.

icambron:
I can't reproduce this in Chrome, where I get `"2019-03-10T04:00:00.000-05:00"`. What browser (or Node version) and what version of Luxon?

kaftand:
ahh sorry for my late response. I am using node v12.14.1 I can try with the current recommended (12.16.1) . here is my script verbatum 
`const lux = require('luxon')

var oneAm = lux.DateTime.local(2019,3,9,3,0,0,0);

console.log(oneAm.plus({days:0, hours:23}).c.hour )
console.log(oneAm.plus({days:0, hours:24}).c.hour )
`
If the first log matches the second log, you have reproduced the issue. Ofcourse, this only works for central time zone.

icambron:
Ah, I was able to reproduce it this time; trick was the `days: 0` part. Yeah, this is a bug.

For future me: issue is almost certainly the `shiftTo` at the top of `adjustTime`, which is collapsing the 24 hours into the `days` key. May have broken in ee70b8b8b2f3c75b863f2fb35e9b8327d5f83eec. Removing the `shiftTo` would break fractional days, since that's how we redistribute the fractional part into the lower-order pieces.

GillesDebunne:
Fixed in #722

PR #722, opened 2020-05-31
Fix for negative values in plus/minus. Fixes #645 and #669

### LINKED ISSUE #645: DateTime.plus(...) with mixed algebraic signs works incorrectly
## Description
If doing a calculation on `DateTime` with mixed algebraic signs leads to incorrect results. This is true for example when modifying dates in a way that the target month does not have 30 days or the target year is a leap-year.

## Example
```
// should be 2019-12-09, but is 2019-12-10
DateTime.fromISO('2020-01-08').plus({months: -1, days: 1})
```

## Analysis
Internally, the duration `{months: -1, days: 1}` is shifted to `months, days, milliseconds` and during that it is normalized. During normalization, the algebraic signs are changed and the resulting duration is `{months: 0, days: -29}`. It would be correct if the target month has 30 days, but December has 31 days.

Same issue for leap-years: `{years: -4, days: 1}` is converted to `{years: -3, days: -364}`, should be `days: -365` if the resulting year is a leap year.

## Reproducible example
https://codesandbox.io/s/sleepy-diffie-x761e

## Version
Tested with latest Luxon version (1.21.3)

## Workaround
Add units one by one in the right order: `DateTime.fromISO('2020-01-08').plus({months: -1}).plus({days: 1})`

GillesDebunne:
Mixed positive and negative duration units values are merged when using `shiftTo`, resulting in unintuitive computations.
 
Removing the shift to milliseconds solves this problem, but then breaks computation with fractional values.

Applying the integer part on large units and convert the fractional part to milliseconds solves this issue.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -122,22 +122,23 @@ function objToTS(obj, offset, zone) {
 
 // create a new DT instance by adding a duration, adjusting for DSTs
 function adjustTime(inst, dur) {
-  const keys = Object.keys(dur.values);
-  if (keys.indexOf("milliseconds") === -1) {
-    keys.push("milliseconds");
-  }
-
-  dur = dur.shiftTo(...keys);
-
   const oPre = inst.o,
-    year = inst.c.year + dur.years,
-    month = inst.c.month + dur.months + dur.quarters * 3,
+    year = inst.c.year + Math.trunc(dur.years),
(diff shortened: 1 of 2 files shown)


icambron:
Can you rebase this? I think it's confused about the earlier commits

Issue #108, opened 2017-12-01
Comparison Methods?


lacivert:
I couldn't find any comparison method like two `DateTime`s, if one is before or after from another.

Momentjs has query functions: http://momentjs.com/docs/#/query/

Will be similar functions in Luxon? 
And what is the best way to compare two DateTime instances for now?

woozyking:
Use natural js comparison operators like `>`, `<=`, etc.

icambron:
Yup @woozyking is right. I'm going to leave this open to figure out where to document those. Probably deserves an entry in the "Math" page of the guide

lacivert:
👍 Thank you very much. I think Math section is good. 

I also think this words would be useful for people who tries 'search' (like Ctrl+F): compare, comparison, after, before

icambron:
Updated the docs in 3918e6b. Will get pushed to the website next time I cut a release.

askerkhakh:
```
+d1 === +d2 // are d1 and d2 the same instant in time?
```
This looks ugly... Like incapsulation violation - knowledge that valueOf returns something useful and unary plus sign for calling it...

icambron:
`valueOf` returning a number that is useful and being involved by the unary `+` is just how JS works. Your complaint is really with JS; Luxon is just following its idioms.

gdebunne:
This is meant as a warning I imagine.

Since `<` and other math comparison operators are available, one may be tempted to use `==` or `===` for equality test. As pointed out, this will usually not work as expected with these immutable objects.

The idiomatic way of testing for equality is probably `d1.equals(d2)`,

or if you are only interested in raw timestamps, `d1.toMillis() === d2.toMillis()`

askerkhakh:
> `valueOf` returning a number that is useful and being involved by the unary `+` is just how JS works. Your complaint is really with JS; Luxon is just following its idioms.

Yep, I don't like the unary plus in general, that's true. I don't like all that silly magic, that JS developers love (double negotiation, boolean "or" in non-boolean context and so on and so forth). Anyway, JS doesn't support operator overloading, so I don't think that using a "hack" with `valueOf` is obvious and expected, otherwise there wouldn't be this issue :wink: 
I wish there were ordinary comparison methods...

delcon:
I recommend outputting and comparing ISO strings (any flavor) like so:
`"2022-04-26T06:00" < "2022-04-26T06:01"`

gdebunne:
[...]

PR #725, opened 2020-06-05
Update wording around date equality test

GillesDebunne:
As discussed in #108

### DIFF
--- a/docs/math.md
+++ b/docs/math.md
@@ -102,7 +102,7 @@ However, beware that `===` compares object identity, which not a useful concept 
 
 ```js
 d1.toMillis() === d2.toMillis() // are d1 and d2 the same instant in time?
-+d1 === +d2 // same test, using object coercision
++d1 === +d2 // same test, using object coercion
 ```
 
 You may also use `#hasSame` to make more subtle comparisons:
--- a/docs/math.md
+++ b/docs/math.md
@@ -98,7 +98,7 @@ DateTime implements `#valueOf` to return the epoch timestamp, so you can compare
 d1 < d2 // is d1 before d2?
 ```
 
-However, beware that `===` compares object identity, which not a useful concept in a library with immutable types. Use `#equals` to compare both the time and additional metadata, such as the locale and time zone. If you're only interested in checking the equality of the timestamps, you can use:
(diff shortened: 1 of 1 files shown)


docs/math.md
```diff
@@ -98,10 +98,11 @@ DateTime implements `#valueOf` to return the epoch timestamp, so you can compare
 d1 < d2 // is d1 before d2?
 ```
 
-However, `===` compares object identity (not a useful concept in a library with immutable types) and `#equals` compares both the time and additional metadata, such as the locale and time zone. If you're only interested in checking the equality of the timestamps, you need to coerce them:
+However, beware that `===` compares object identity, which not a useful concept in a library with immutable types. Use `#equals` to compare both the time and additional metadata, such as the locale and time zone. If you're only interested in checking the equality of the timestamps, you can use:
 
 ```js
-+d1 === +d2 // are d1 and d2 the same instant in time?
+d1.toMillis() === d2.toMillis() // are d1 and d2 the same instant in time?
++d1 === +d2 // same test, using object coercision
```
icambron, line 105: coercion

docs/math.md
```diff
@@ -98,10 +98,11 @@ DateTime implements `#valueOf` to return the epoch timestamp, so you can compare
 d1 < d2 // is d1 before d2?
 ```
 
-However, `===` compares object identity (not a useful concept in a library with immutable types) and `#equals` compares both the time and additional metadata, such as the locale and time zone. If you're only interested in checking the equality of the timestamps, you need to coerce them:
+However, beware that `===` compares object identity, which not a useful concept in a library with immutable types. Use `#equals` to compare both the time and additional metadata, such as the locale and time zone. If you're only interested in checking the equality of the timestamps, you can use:
```
icambron, line 101: which is not

icambron, line 101: Also you want "be aware that"; "beware that" doesn't sound quite right

GillesDebunne, line 101:
```suggestion
However, be aware that `===` compares object identity, which is not a useful concept in a library with immutable types. Use `#equals` to compare both the time and additional metadata, such as the locale and time zone. If you're only interested in checking the equality of the timestamps, you can use:
```

docs/math.md
GillesDebunne, line 105:
```suggestion
+d1 === +d2 // same test, using object coercion
```

PR #732, opened 2020-07-02
Fixed typos in docs where 'month' was used in place of 'weekday'

jonknowles:
This is a tiny update to the Info docs that fixes places where 'month' was used in place of 'weekday' in the annotations for `Info.weekdays()` and `Info.weekdaysFormat()`.


icambron:
Thanks!

PR #737, opened 2020-07-09
feat: add support for large durations in seconds

### LINKED ISSUE #731: ISO Duration with 10 digits seconds cannot be parsed as ISO 8601
### Steps to reproduce

```js
const duration = luxon.Duration.fromISO('PT1000000000S')
console.log(duration.invalid)
```

### Current Result

```
Object {
  explanation: "the input 'PT1000000000S' can't be parsed as ISO 8601",
  reason: "unparsable",
  toMessage: function(){return this.explanation?this.reason+": "+this.explanation:this.reason}
}
```

### Expected Result

```js
null
```

### Details

`Duration.fromISO(isoDurationString)` cannot parse an ISO Duration that has seconds with 10 or more leading digits (eg. 'PT1000000000S').

Unfortunately this is the kind of format that `DateTime.fromISO(isoDateString).diffNow().negate().toISO()` produces.

See this Codepen for a minimal example:

https://codepen.io/visualjerk/pen/rNxpVBW

visualjerk:
Closes #731


icambron:
Looks good. I'll release this very soon.

PR #758, opened 2020-08-16
Support arbitrary precision in ISO milliseconds. Fixes #757

### LINKED ISSUE #757: fromISO with 10 digit milliseconds
Seems when parse milliseconds long than 9 decimals its returned as invalid.
**good**: `2020-04-03T11:41:19.539000000`
**invalid**: `2020-04-03T11:41:19.5390000001`
```
var date = DateTime.fromISO('2020-04-03T11:41:19.5390000001')
// returns false
console.log(date.invalid)
```
[jsFiddle Here](https://jsfiddle.net/hxe5762t/1/)

_Originally posted by @cmcleese in https://github.com/moment/luxon/issues/632#issuecomment-608405403_

GillesDebunne:
Simply remove the 9 digit limit in the regexp

### DIFF
--- a/src/impl/regexParser.js
+++ b/src/impl/regexParser.js
@@ -66,7 +66,7 @@ function simpleParse(...keys) {
 
 // ISO and SQL parsing
 const offsetRegex = /(?:(Z)|([+-]\d\d)(?::?(\d\d))?)/,
-  isoTimeBaseRegex = /(\d\d)(?::?(\d\d)(?::?(\d\d)(?:[.,](\d{1,}))?)?)?/,
+  isoTimeBaseRegex = /(\d\d)(?::?(\d\d)(?::?(\d\d)(?:[.,](\d{1,30}))?)?)?/,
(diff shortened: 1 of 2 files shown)


src/impl/regexParser.js
```diff
@@ -66,7 +66,7 @@ function simpleParse(...keys) {
 
 // ISO and SQL parsing
 const offsetRegex = /(?:(Z)|([+-]\d\d)(?::?(\d\d))?)/,
-  isoTimeBaseRegex = /(\d\d)(?::?(\d\d)(?::?(\d\d)(?:[.,](\d{1,9}))?)?)?/,
+  isoTimeBaseRegex = /(\d\d)(?::?(\d\d)(?::?(\d\d)(?:[.,](\d{1,}))?)?)?/,
```
icambron, line 69: We can't make it arbitrary, because that opens up the library to regex DOS scenarios in Node contexts. We can make it like `20` though

PR #777, opened 2020-09-18
Update README.md

rqx110:
correct the sample code


GillesDebunne:
The sample code is actually correct.
`now` is a new method introduced in commit cf6e87799b107845c6c2e0d257caf0fc31ab5a41 which is simply `local()` with no parameter.

PR #785, opened 2020-09-24
UnitFormat was folded into Unified NumberFormat

benkelaar:
Hiya, was looking into proper internationalisation of durations and encountered this note in the luxon docs.

As per https://github.com/tc39/ecma402/issues/32#issuecomment-516588266, UnitFormat was merged into Unified NumberFormat, which was merged into ECMA-402 last March (2020).

Now I haven't investigated how widely it's already supported by browsers but I think this update of the documentation is at least warranted.

Should I create an issue to start preparing for a `Duration.humanize`-like method?


GillesDebunne (approved):

benkelaar:
@GillesDebunne Thanks!  Do you think I should create an issue for `Duration.humanize`?

GillesDebunne:
If it is not yet available in browsers, this issue might be as good as a reference.

Issue #255, opened 2018-05-10
DateTime instance from javascript Date instance


damianobarbati:
Is it possible to convert a native Date instance to DateTime instance?
Something like.
```
const date = new Date(); //somehow I have a date instance
const datetime = DateTime.fromDate(date); //I want to convert it to a Luxon DateTime instance
```

icambron:
https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#static-method-fromJSDate

xinthose:
I do not see `fromJSDate` anywhere in the manual.  I had to search google to find this.

GillesDebunne:
Indeed, it should probably be part of `docs/manual/parsing.html`. Added in #789

mayrsascha:
I was looking for this on Google even after you guys added it to the parsing docs page. IMO it would be nice to have this on the "A quick tour" page [where you also describe how to create a `DateTime` from an ISO8601 String](https://moment.github.io/luxon/docs/manual/tour.html#parse-from-iso-8601), as it seems to be a common use case to me.

revington:
all links are broken

AwesomeDevDen:
@revington https://moment.github.io/luxon/api-docs/index.html

revington:
thx @AwesomeDevDen

LarsFlieger:
It has to be:
```ts
const date = new Date();
const datetime = DateTime.fromJSDate(date); 
```
Docs: https://moment.github.io/luxon/#/parsing?id=js-date-object

srwaggon:
This is such a common usecase that I can't believe it isn't featured at the top of the "A quick tour/Creating a DateTime" section.

PR #789, opened 2020-09-30
Add fromJSDate reference in parsing.md #255

GillesDebunne:


docs/parsing.md
```diff
@@ -81,6 +81,13 @@ DateTime.fromSeconds(1542674993);
 
 Both methods accept the same options, which allow you to specify a timezone, calendar, and/or numbering system.
 
+### JS Date Object
+
+A native JS `Date` object can be converted into a `DateTime` using [fromJSDate](../class/src/datetime.js~DateTime.html#static-method-fromJSDate).
+
+An optional `zone` option can be provided for accurate DST date computations.
```
icambron, line 88: "Accurate" is not quite the right word, since it implies they'd be inaccurate if the parameter was not provided. "zone-specific" would be better, but IMO even better would be "An optional `zone` parameter can be provided to set the zone on the resulting object", which covers all the bases

icambron (approved):

PR #797, opened 2020-10-03
Fix ES6 minify

### LINKED ISSUE #771: URL broken in install instructions
It links to here which is 404: `https://moment.github.io/luxon/es6/luxon.min.js`

https://moment.github.io/luxon/docs/manual/install.html#es6

pgcalixto:
This is an attempt to resolve #771.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "luxon",
-  "version": "1.24.1",
+  "version": "1.25.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


pgcalixto:
@icambron, how is the broken URL re-generated (https://moment.github.io/luxon/es6/luxon.min.js)?

pgcalixto:
Also, I couldn't help but notice that the package-lock still had the previous version (you can see in the diff: `"version": "1.24.1"`).

It would be ideal for it to match the version in package.json. Is there an automated process to bump them both and keep them matching?

icambron:
@pgcalixto I'm not sure. I'll figure that out when I do the next release, which should be very soon

icambron:
Thanks!

PR #798, opened 2020-10-04
Fix hasSame unit comparison

### LINKED ISSUE #584: hasSame order matters / timezones are not working with hasSame
I noticed that when I compared two DateTime objects with hasSame the ordering of the objects returned different results. It seems to only happen when at least one of the objects has a timezone defined.

I managed to reproduce it, let the code speak for itself:

```
// With zones
const first = DateTime.fromISO("2019-10-02T07:24:04.339+03:00", {
  zone: "Europe/Helsinki"
});
const second = DateTime.fromISO("2019-10-02T00:00:00.000-05:00", {
  zone: "America/Chicago"
});

console.log(first.hasSame(second, "day")); // true
console.log(second.hasSame(first, "day")); // false
console.log(first.toISO()); // 2019-10-02T07:24:04.339+03:00
console.log(second.toISO()); // 2019-10-02T00:00:00.000-05:00

// No zones
const third = DateTime.fromISO("2019-10-02T07:24:04.339+03:00");
const fourth = DateTime.fromISO("2019-10-02T00:00:00.000-05:00");

console.log(third.hasSame(fourth, "day")); // true
console.log(fourth.hasSame(third, "day")); // true
console.log(first.toISO()); // 2019-10-02T07:24:04.339+03:00
console.log(second.toISO()); // 2019-10-02T08:00:00.000+03:00
```

danilotitato:
This is an attempt to resolve #584

### DIFF
--- a/docs/math.md
+++ b/docs/math.md
@@ -108,9 +108,8 @@ d1.toMillis() === d2.toMillis() // are d1 and d2 the same instant in time?
 You may also use `#hasSame` to make more subtle comparisons:
 
 ```js
-d1.hasSame(d2, 'millisecond'); // equivalent to `+d1 === +d2`
-d1.hasSame(d2, 'minute');       // both DateTimes are in the same minute (and hour, day, month, etc)
-d1.hasSame(d2, 'year');         // etc
+d1.hasSame(d2, 'year');   // both DateTimes have the same calendar year
+d1.hasSame(d2, 'day');    // both DateTimes have the same calendar day (which implies they also have the same calendar year and month)
 ```
 
 Note that these are checking against the calendar. For example, if `d1` is in 2017, calling `hasSame` with "year" asks if d2 is also in 2017, not whether the DateTimes within a year of each other. For that, you'd need `diff` (see below).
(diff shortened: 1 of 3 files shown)


danilotitato:
@icambron I've based this fix on your comment on rezoning the other DateTime (https://github.com/moment/luxon/issues/584#issuecomment-541160618).
Is this what you were expecting?

If so, what test cases do you believe it would be needed to add?
From the current ones, none of them would check the mentioned issue.

GillesDebunne:
Kudos for using `keepLocalTime` which effectively compares *local* calendar times.

danilotitato:
@GillesDebunne thank you for your input. I've merged your PR on this one but seems that codecov didn't liked it. Could you help me with it? I can't tell for sure if the test coverage has changed and why.

Also, do you guys think this would be a breaking change (which would lead to a bump in the major version)? Maybe some people were using this assimetrically. Or it can be considered just a bugfix?

GillesDebunne:
The code coverage warning can probably be ignored, it comes from new lines in the documentation that affect the ratio.

Other similar bugs were fixed before without bumping the major version number.

icambron:
Sorry for taking so long to review this. It looks good, thanks!

PR #800, opened 2020-10-04
Bump version with npm script

### LINKED ISSUE #799: package-lock.json's version does not match package.json's
It would be ideal to have the version on `package-lock.json` matching `package.json`'s.
It seems to be missing as a part of the release creation.

- https://github.com/moment/luxon/blob/345b1cdca1705af1bc4846158e8dcde85febd7fb/package.json#L3
- https://github.com/moment/luxon/blob/345b1cdca1705af1bc4846158e8dcde85febd7fb/package-lock.json#L3
- https://github.com/moment/luxon/commit/345b1cdca1705af1bc4846158e8dcde85febd7fb

pgcalixto:
This is a suggestion to resolve #799.

This is not necessarily intended to be merged. It was created mainly as a suggestion for further discussion and improvements.

### DIFF
--- a/scripts/tag
+++ b/scripts/tag
@@ -1,10 +1,5 @@
 #!/usr/bin/env bash
 
-PACKAGE_VERSION=$(node -p "require('./package.json').version")
+npm version ${1:-minor} -m "bump to %s"
 
-if [ $(git tag -l "$PACKAGE_VERSION") ]; then
-  echo "tag already exists"
-else
-  git tag $PACKAGE_VERSION
-  git push origin master $PACKAGE_VERSION 
-fi
+git push origin master


pgcalixto:
Changed it to draft since it is more of a suggestion up for discussion than a PR ready to merge.

GillesDebunne:
I believe the bump version process is manual. It requires writing a summarized changelog, as well as deciding between a minor and a patch release, and updating `package.json` accordingly.

I think `VERSION` should be a hardcoded constant exported in `luxon.js` along with the other objects. That would be an other manual step when bumping version, but it is not that common.

`script/release` should however ensure the consistency, using `require(../package.json).version` as a reference, and checking it matches with `package-lock` and `VERSION`. It should abort if that is not the case, and could provide hints on the commands needed to update the latest commit accordingly (applying those automatically seems too bold).

What do you think?

pgcalixto:
I agree with the hardcoded `VERSION`, while making `script/release` ensure the consistency.

The approach you suggested makes total sense. I will try to develop this approach in the other PR (https://github.com/moment/luxon/pull/794), then I'll close this one.

pgcalixto:
Closing this as it is being reworked on https://github.com/moment/luxon/pull/794.

pgcalixto:
Closing this as it is being reworked on #794.

Issue #625, opened 2019-12-01
Add Duration.fromHTMLTime()


munierujp:
Value of HTML's `<input type="time">` have below patterns:

- `hh:mm`
- `hh:mm:ss`
- `hh:mm:ss.S`
- `hh:mm:ss.SS`
- `hh:mm:ss.SSS`

These are expressed as below regular expression:

```
/^(\d{2}):(\d{2})(:(\d{2}))?(\.(\d{1,3}))?$/
```

cf. https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#times

## My idea
*Not tested*

### Implemention

```js
function fromHTMLTime (text) {
  const matched = text.match(/^(\d{2}):(\d{2})(:(\d{2}))?(\.(\d{1,3}))?$/)

  if (!matched) {
    return Duration.invalid('unparsable', `the input "${text}" can't be parsed as HTML time`)
  }

  const getValue = index => matched[index] ? Number(matched[index]) : 0
  const hours = getValue(1)
  const minutes = getValue(2)
  const seconds = getValue(4)
  const milliseconds = getValue(6)
  return Duration.fromObject({
    hours,
    minutes,
    seconds,
    milliseconds
  })
}
```

### Usage example

```html
<input type="time" id="time" value="13:30">
```

```js
const timeElement = document.getElementById('time')
const value = timeElement.value  // "13:30"
const duration = Duration.fromHTMLTime(value)
```

icambron:
A couple thoughts on this:

 * It's not clear to me that HTML time is meant to specify a duration, as opposed to a specific time on some (externally-provided) date, i.e. it seems like `13:30` means the time, as opposed to a period of 13 hours and 30 minutes.
 * As a *time*, it's just an ISO time and the `fromISO` method handles, though of course it returns a `DateTime`, not a `Duration`

I'm not against providing ISO-like parsing for durations, I just don't think it really has much to do with HTML Time.

icambron:
Let's do this:

```
Duration.fromISOTime("13:30");
Duration.fromISOTime("13:30:30");
Duration.fromISOTime("13:30:30.1");
Duration.fromISOTime("13:30:30.12");

thomas-darling:
@icambron We'd _really_ like this too. Do you have any idea when support for this might land?

[...]

PR #803, opened 2020-10-06
Add fromISOTime, toISOTime and toMillis to Duration

thomas-darling:
This resolves https://github.com/moment/luxon/issues/625, by adding a `fromISOTime` and `toISOTime` method to the `Duration` class.
The design of the methods is closely aligned with the design of similar, existing methods, such as `fromISO` and `toISO`.

Additionally, since the `Duration` class has a `fromMillis` method, I've also added a corresponding `toMillis` method.
The justification for this is, that having corresponding pairs of from/to methods, makes it easy to roundtrip a value in the same format, and that it aligns with the `DateTime` class, which already has both method.


src/duration.js
```diff
@@ -432,6 +457,57 @@ export default class Duration {
...
+      value.milliseconds > 0 || !opts.suppressMilliseconds ? "hh:mm:ss.SSS" :
+      value.seconds > 0 || !opts.suppressSeconds ? "hh:mm:ss" :
+      value.minutes > 0 || !opts.suppressMinutes ? "hh:mm" :
+      "hh";
+
+    if (opts.format !== "extended") {
+      format = format.replace(/:/g, "");
+    }
+
+    if (opts.suppressPrefix === "never" || opts.suppressPrefix === "auto" && (format === "hh" || format === "hhmm")) {
```
icambron, line 505: I'm not against options like these, but we need to make sure they're consistent. We don't support these elsewhere, do we?

src/impl/regexParser.js
```diff
@@ -118,6 +118,29 @@ function extractIANAZone(match, cursor) {
   return [{}, zone, cursor + 1];
 }
 
+// ISO time parsing
+
+const isoTimeOnly = /^T?(\d\d)(?::?(\d\d)(?::?(\d\d)(?:[.,](\d{1,30}))?)?)?$/;
```
icambron, line 123: Aren't these identical to `isoTimeBaseRegex` and `extractISOTime`?

src/duration.js
thomas-darling: Yep, `DateTime` has similar options, so I designed this to behave the same way.
See https://github.com/moment/luxon/blob/master/src/datetime.js#L1534

src/impl/regexParser.js
thomas-darling: They are very similar, but not quite identical.

* `isoTimeOnly` includes the optional `T` prefix, and matches the beginning/end of the string - `isoTimeBaseRegex` does not.

  This means `isoTimeBaseRegex` would match a value such as `foo00bar`, even though that's obviously not a valid Time value.
  But I guess we could consider changing it to this, to reuse the existing regex:

  ```
  const isoTimeOnly = RegExp(`^T?${isoTimeBaseRegex.source}$`)
  ```

* `extractISOTimeOnly` preserves the _precision_ of the value specified in the string, by only setting the values actually specified in the string - `isoTimeBaseRegex` always sets e.g. `seconds` to 0, even if not specified.

  > Either the seconds, or the minutes and seconds, may be omitted from the basic or extended time formats for greater brevity but decreased precision - https://en.wikipedia.org/wiki/ISO_8601#Times

  One could of course argue `extractISOTime` should be fixed so it too preserves the precision - I'm just afraid that might be a breaking change, as code may have been written to expect `seconds` in a `DateTime` instance to always have a value.

  There's also a minor difference in the casing of the property names in the returned object - e.g. `seconds` instead of `second`. That's because `Duration.fromObject` only accepts the plural version, which kinda makes sense, as it may represent arbitrary durations, and the singular name only makes sense when talking specifically about a time of day.

src/duration.js
thomas-darling: Specifically regarding the prefix suppression - this is unique to the time format:

> As of ISO 8601-1:2019, the basic format is T[hh][mm][ss] and the extended format is T[hh]:[mm]:[ss]. Earlier versions omitted the T in both formats.
> ...
> ISO 8601-1:2019 allows the T to be omitted in the extended format, as in "13:47:30", but only allows the T to be omitted in the basic format when there is no risk of ambiguity with date expressions.
>
> https://en.wikipedia.org/wiki/ISO_8601#Times

So I made the default behavior here `auto`, i.e. omit if possible - but if you're interacting with a system that always/never expects the prefix, this option allows you to override that behavior.

thomas-darling:
Hmm, looking at this again, I'd actually like to refactor it a little.
The way I implemented the options here, is not quite as consistent with the rest of the code as I thought it was 🙂

icambron:
@thomas-darling ok, let me know when you want me to take another look. I can spend some real time looking at this when it's ready; my comments so far were the result of only a very quick readthrough

thomas-darling:
@icambron, will do 👍
I got a little busy here, but I'm working on it, so should have it ready for review soon 🙂

test/datetime/format.test.js
```diff
@@ -133,10 +133,6 @@ test("DateTime#toISOTime({suppressSeconds: true}) will suppress milliseconds if
   expect(dt.set({ second: 0, millisecond: 0 }).toISOTime({ suppressSeconds: true })).toBe("09:23Z");
 });
 
-test("DateTime#toISOTime({suppressSeconds: true}) will suppress milliseconds if they're zero", () => {
```
thomas-darling, line 136: Removed because it is identical to the test above

test/datetime/format.test.js
```diff
@@ -133,10 +133,6 @@ test("DateTime#toISOTime({suppressSeconds: true}) will suppress milliseconds if
...
-  expect(dt.set({ second: 0, millisecond: 0 }).toISOTime({ suppressSeconds: true })).toBe("09:23Z");
-});
-
```
thomas-darling, line 139: Removed because it is identical to the test above

test/duration/parse.test.js
```diff
@@ -74,3 +74,36 @@ test("Duration.fromISO rejects junk", () => {
...
+});
+
+test("Duration.fromISOTime can parse a variety of basic ISO time formats", () => {
+  checkTime("112233.444", { hours: 11, minutes: 22, seconds: 33, milliseconds: 444 });
+  checkTime("112233", { hours: 11, minutes: 22, seconds: 33 });
+  checkTime("1122", { hours: 11, minutes: 22, seconds: 0 });
+  checkTime("11", { hours: 11, minutes: 0, seconds: 0 });
+  checkTime("T1122", { hours: 11, minutes: 22, seconds: 0 });
+});
+
```
thomas-darling, line 100: This exactly matches how the time component of a date is parsed - including the, in my opinion, strange behavior where milliseconds are undefined if not specified, while the other values default to 0.

A general discussion could be had about whether defaulting to 0 is the correct behavior, as it means we actually loose the information about the _precision_ of the value. That said, the need for that information is probably an uncommon edge case, and defaulting to 0 is very convenient, as it means users won't need null checks - so I'm ok with that.

However, I do think milliseconds should then also default to 0, both here and when parsing a date, as the current behavior is quite unexpected - especially since the typings for DateTime claim the milliseconds will always be a number...

Thoughts?

src/duration.js
```diff
@@ -433,6 +458,53 @@ export default class Duration {
...
+      }
+    }
+
+    if (opts.includePrefix) {
+      fmt = "T" + fmt;
+    }
+
+    return value.toFormat(fmt);
+  }
+
```
thomas-darling, line 507: This method is designed to behave just like the existing `toISOTime()` in the `DateTime` class - and that one has been extended to support adding the prefix, but is otherwise unchanged.

In my first implementation, the default was to automatically choose whether to add the prefix, depending on the format. However, I've since decided to drop that, and instead make it an explicit option, defaulting to false.

Technically, the 2019 version of the spec does require the prefix to be present for the basic format `hhmm`, but the previous version didn't, and changing it now would therefore have been a breaking change.

Also, since the user already has to specify options to get the basic format, they might as well also specify if they want the prefix...

thomas-darling:
@icambron I think this is done now, so feel free to review when you have the time 🙂 

(by the way, I think the CI build may be stuck - looks like it's been queued for quite a while now)

thomas-darling:
@icambron, turns out I had a few bugs there - but should be all done now 🙂

JensUngerer:
I've justed "toied with the idea of doing" a merge request for a ``toMillis`` method in the Duration class... .
I am happy that i found this merge - request a minute ago...

I am using ``luxon`` for my [TimeTracker](https://github.com/JensUngerer/time-tracker) project.
It is exactly what i've searched for.
"But", it "lacked" one kind of "functionality": a ``toMillis`` method in the Duration class... .

> @icambron, turns out I had a few bugs there - but should be all done now 🙂
I am looking forward seeing the merge - request to be merged.

icambron:
Sorry for the long delay. Merged and will be released shortly. Thanks!

PR #820, opened 2020-10-23
Fix typo in token formatting docs

vaicine:
The t token does not include the meridiem


icambron:
Ah, not a typo. Whether it's there or not is just part of the localization. For me, in en-US:

```js
DateTime.local().toFormat("t"); //=> "9:45 AM"

// but

DateTime.local().setLocale("en-gb").toFormat("t"); //=> "09:45"
```

PR #824, opened 2020-10-28
Stick to ES5 functions

tjenkinson:
This adds a few more helpers so that the build is not using ES6 functions (which are undetectable from static analysis).

We would like to use Luxon on a platform where we have only polyfilled up to ES5, and it's risky for us to add more polyfills. We can run the output through the typescript compiler with a target of ES5 though, so things like `const`, `let`, `[...something]` etc are ok.

I noticed there are already helpers for other ES6 things like `padStart` and `padEnd`, so I'm hoping this is ok?

Thanks!

### DIFF
--- a/src/impl/util.js
+++ b/src/impl/util.js
@@ -201,6 +201,22 @@ export function findIndex(array, predicate) {
   return -1;
 }
 
+export function is(x, y) {
+  if (typeof Object.is === "function") {
+    return Object.is(x, y);
+  }
+  // SameValue algorithm
+  if (x === y) {
+    // Steps 1-5, 7-10
+    // Steps 6.b-6.e: +0 != -0
+    return x !== 0 || 1 / x === 1 / y;
+  } else {
+    // Step 6.a: NaN == NaN
+    // eslint-disable-next-line no-self-compare
+    return x !== x && y !== y;
+  }
+}
+
 // NUMBERS AND STRINGS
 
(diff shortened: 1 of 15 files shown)


src/impl/util.js
```diff
@@ -87,6 +87,108 @@ export function hasOwnProperty(obj, prop) {
   return Object.prototype.hasOwnProperty.call(obj, prop);
 }
 
+export function assign(target /* , ...sources */) {
+  if (target === null || target === undefined) {
```
tjenkinson, line 91: this and the others pretty much copied/pasted from mdn

gdebunne:
If I understand correctly, you are replacing ES6 built-in functions with ES5 polyfills. What about performance issues on recent browsers, where these native methods could have an optimized implementation? What's the impact on bundle size?

Also, how to make sure future developments do not break this compatibility?

GillesDebunne:
@icambron I let you comment on this proposal...

icambron:
This is a lot of code to add for stuff that's already built into the vast majority of the JS environments, and I'm not going to bloat the library because you don't want to polyfill your application; that's not a good tradeoff between your needs and the weight of the library.

The way to support this use case is to define an ES6 object that exports these methods, with trivial ES6 implementations. Then you can `set` that object with your own implementation in.

GillesDebunne:
Also, is there no way to transpile from ES6 to ES5, so that no polyfill is required ?

PR #841, opened 2020-12-11
Fixed broken link to cldr.unicode.org

bmispelon:
I noticed that the link to http://cldr.unicode.org/translation/date-time was broken (404).

I believe the new URL is http://cldr.unicode.org/translation/date-time-1/date-time.

Seeing as I didn't change any code (only a comment) I didn't run the tests but I did run the linter.


icambron:
Thanks

PR #855, opened 2021-01-15
Remove profanity from comments

mj1856:
I appreciate the sentiment, but this causes adoption problems in projects that scan for sensitive terminology.


icambron:
lol @ potty language scanners for source code.

But sure, no problem

Issue #859, opened 2021-01-26
DateTime.set({year, weekNumber}) returns DateTime with current year always


rolandjohann:
**Describe the bug**
Setting DateTime `year=2020` and `weekOfYear=22` leads to a DateTime at year `2021`. Setting `year` and `weekNumber` independently workarounds this issue.

**To Reproduce**
```js
luxon.DateTime.local().set({year: 2020, weekNumber: 22}).toString()
```

**Actual vs Expected behavior**
```js
luxon.DateTime.local().toString()
> "2021-01-26T11:22:24.460+01:00"

luxon.DateTime.local().set({year: 2020, weekNumber: 22}).toString()
> "2021-06-01T11:22:43.287+02:00"

luxon.DateTime.local().set({year: 2020}).set({weekNumber: 22}).toString()
> "2020-05-31T11:23:01.104+02:00"
```

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Chrome 84
- Luxon version: current from official Luxon docs

icambron:
This is a bug, but the bug is that Luxon shouldn't really let you do that. What you almost certainly want to say is:

```js
luxon.DateTime.local().set({weekYear: 2020, weekNumber: 22}).toString()
```

With `22` it makes no difference, but with `1` or `52` it matters a great deal.

Luxon typically prevents you from mixing incompatible units. There should be a check for `year` here: https://github.com/moment/luxon/blob/master/src/datetime.js#L667

mateusbw:
Helo, i just opened this [PR](https://github.com/moment/luxon/pull/861) that i think fixes this issue.

patxbert:
I had a similar problem where I did not realise I should be using `weekYear` instead of `year` for create a new `DateTime` using `DateTime.fromObject()`

**What I thought I needed to do:** 
`DateTime.fromObject({year: 2020, weekNumber: 10})`
**What I was doing:**
`DateTime.fromObject({year: 2020}).set({weekNumber: 10})`
**What I should have done:**
`DateTime.fromObject({weekYear: 2020, weekNumber: 10})`

saul-gush:
Just tried this example from the documentation and the hour doesn't change to 3. 
``` javascript
var dt = DateTime.now();
 dt.set({ hour: 3 });
```
Output: 2021-06-28T20:19:49.044+01:00

My sandbox: https://codesandbox.io/s/datetime-formatting-luxon-w5gyf?file=/src/App.js:1432-1481

[...]

PR #861, opened 2021-01-31
#859: Adding validation for mixing incompatible units

mateusbw:
### Problem:
Luxon wasn't preventing from mixing incompatible units.

### Solution
Added validation.


icambron:
Thanks. I will take a good look at this shortly.

PR #867, opened 2021-02-10
Remove fixed-offset formatting workaround #866

### LINKED ISSUE #866: Remove fixed-offset TZ formatting workaround
The workaround I'm talking about: https://github.com/moment/luxon/blob/6934632be3d1003b48a4f43178d6ac1f36bd6911/src/impl/locale.js#L186-L200

I've quickly tried what I've proposed in https://github.com/moment/luxon/issues/201#issuecomment-776784273 and it seems to be working, so I'm opening this issue.

The problem is there are no UTC+X entries in `tzdata` and there are only whole hours from `Etc/GMT-14` to `Etc/GMT+12`.

Here is the list of time zones under `Etc/` on my Linux:

```
[jakub@localhost ~]$ ls /usr/share/zoneinfo/Etc/
GMT    GMT+1   GMT+11  GMT-13  GMT+3  GMT+5  GMT+7  GMT+9      Universal
GMT+0  GMT-1   GMT-11  GMT-14  GMT-3  GMT-5  GMT-7  GMT-9      UTC
GMT-0  GMT+10  GMT+12  GMT+2   GMT+4  GMT+6  GMT+8  Greenwich  Zulu
GMT0   GMT-10  GMT-12  GMT-2   GMT-4  GMT-6  GMT-8  UCT
```

A "good enough" solution could be to still fall back to that workaround for anything outside of the supported range?
EDIT: I've opened #867 that does exactly that.

This should fix issues like #201 and #834

durasj:
Fixes #866

### DIFF
--- a/src/impl/locale.js
+++ b/src/impl/locale.js
@@ -183,20 +183,28 @@ class PolyDateFormatter {
 
     let z;
     if (dt.zone.universal && this.hasIntl) {
-      // Chromium doesn't support fixed-offset zones like Etc/GMT+8 in its formatter,
-      // See https://bugs.chromium.org/p/chromium/issues/detail?id=364374.
-      // So we have to make do. Two cases:
-      // 1. The format options tell us to show the zone. We can't do that, so the best
-      // we can do is format the date in UTC.
-      // 2. The format options don't tell us to show the zone. Then we can adjust them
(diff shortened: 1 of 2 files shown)


icambron:
Looks great

Issue #775, opened 2020-09-16
weekdayLong is slow to execute


rooch84:
I'm finding that `weekdayLong` is approximately two orders of magnitude slower that `weekday`.  This is especially apparent when processing large volumes of csv data. 

Luxon version is 1.25.0

From a quick look at the code, I would guess it's because a Locale object is being created for each call. 

```js
const {DateTime } = require("luxon")

const NUM_TO_DAY = {
  1: "Monday",
  2: "Tuesday",
  3: "Wednesday",
  4: "Thursday",
  5: "Friday",
  6: "Saturday",
  7: "Sunday",
};

const date = DateTime.local();

console.time("weekdayLong");
const weekdayLong = date.weekdayLong;
console.log(weekdayLong);
console.timeEnd("weekdayLong");


console.time("weekdayLookup");
const weekdayLookup = NUM_TO_DAY[date.weekday];
console.log(weekdayLookup);
console.timeEnd("weekdayLookup");
```

```bash
Wednesday
weekdayLong: 5.281ms
Wednesday
weekdayLookup: 0.039ms
```

icambron:
It's probably because it's using the Intl API, which is hella slow. For _most_ English operations, we shortcutted that with hardcoded implementations. I wonder if that's not happening in this case...

dan-overton:
weekdayLong is looking up from hardcoded data, but the Locale create does make it slower. Using the existing loc object from the `DateTime` is roughly twice as quick over 100000 iterations.

PR #877, opened 2021-02-21
Add opt to use existing loc obj in weekdays/months

dan-overton:
Potential fix for https://github.com/moment/luxon/issues/775 .

This allows an existing loc object to be passed into Info `months` and `weekdays` functions, to be used instead of creating a new one. The use of these in DateTime has been modified to use this option.

In a loop of 100000 uses of `DateTime.weekdayLong` this appears to be roughly twice as fast as creating a Locale each time.

### DIFF
--- a/benchmarks/datetime.js
+++ b/benchmarks/datetime.js
@@ -4,64 +4,76 @@ import Benchmark from "benchmark";
 import DateTime from "../src/datetime";
 import Settings from "../src/settings";
 
-const suite = new Benchmark.Suite();
+function runDateTimeSuite() {
+  return new Promise((resolve, reject) => {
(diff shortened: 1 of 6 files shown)


icambron:
Good stuff. Can you add something to the benchmarks code?

dan-overton:
Sure, will do

dan-overton:
I've added some benchmarks. The benchmarks needed to be at the Info level to compare running with / without an existing locale object. That led to me refactoring the benchmarks themselves to run multiple files. Let me know what you think  - I'm happy to roll it back a bit if there's a simpler way that works.

icambron:
Seems fine. Thanks!

Issue #829, opened 2020-11-05
Info.weekdays seems to be mutable


UnsungHero97:
**Describe the bug**
It seems that `Info.weekdays` is mutable and it affects `DateTime` instances.

**To Reproduce**
To reproduce, I just called `.pop()` on `Info.weekdays` in the console on your website:

https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html

Here's a screenshot of the console:

![Screen Shot 2020-11-05 at 10 59 59 AM](https://user-images.githubusercontent.com/440299/98284515-154dc980-1f56-11eb-807a-9ee8b991c123.png)

**Actual vs Expected behavior**
Luxon claims to be immutable. My expectation is that running `Info.weekdays('short')` would give me a new copy of the array instead of a global constant that is used by `DateTime` instances.

**Desktop (please complete the following information):**

- macOS
- Chrome 86
- luxon v1.25.0

**Additional context**
The reason I am doing this is because I'm trying to get a copy of `Info.weekdays('short')` but with Sunday being at the beginning of the array.

icambron:
Yeha, Luxon should copy the array before returning it. Would accept a PR

UnsungHero97:
Thanks, @icambron. I'll put together a PR.

UnsungHero97:
Sorry y'all, haven't had time to work on that pull request.

dan-overton:
I had some time so opened a PR to fix this, any review would be appreciated.

icambron:
Closed by #829

PR #886, opened 2021-02-27
#829: Made English Info lists immutable

dan-overton:
Fix for [#829](https://github.com/moment/luxon/issues/829).

Bug only appears when the English lists are used, hence setting the default locale in the test.


icambron:
Look good, thanks!

Issue #856, opened 2021-01-18
Setting zone to Etc/GMT0 yields "Invalid DateTime"


dmalan:
**Describe the bug**
Setting the time zone of a `DateTime` to `Etc/GMT0` yields "Invalid DateTime."

**To Reproduce**
```js
>> luxon.DateTime.local().setZone('Etc/GMT-0').toString()
"2021-01-18T05:43:02.350Z"
>> luxon.DateTime.local().setZone('Etc/GMT0').toString()
"Invalid DateTime"
```

**Actual vs Expected behavior**
Expected identical behavior from both, since `Etc/GMT0` and `Etc/GMT-0` (and `Etc/GMT+0`) are aliases for `Etc/GMT` as of version [2020f](https://www.iana.org/time-zones) of the tz database.

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Chrome 87, Safari 14, Firefox 84
- Luxon version 1.25.0

**Additional context**
We noticed because we're populating a drop-down of time zones using Moment's [`moment.tz.names()`](https://momentjs.com/timezone/docs/#/data-loading/getting-zone-names/) function, which includes `Etc/GMT0`. Thought it might just browser-specific, but it seems to happen on Chrome, Safari, and Firefox alike?

Many thanks!

icambron:
Yeah, this can probably be fixed, because those are parsed explicitly by Luxon. I suspect we can just amend this line: https://github.com/moment/luxon/blob/master/src/zones/IANAZone.js#L112

dmalan:
Thanks very much!

PR #897, opened 2021-03-13
856: fixed Etc/GMT0 yields Invalid DateTime

raguljr:
Fixed [Issue 856](https://github.com/moment/luxon/issues/856)


raguljr:
Hi @icambron  Can we review & close this bug?

Issue #905, opened 2021-03-22
[Uncaught RangeError: Invalid time value] when passing very high value to zoned DateTime.plus()


gumgl:
**Describe the bug**
I'm not sure if this is out of [spec](https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-plus), but I think this exception should not be thrown. Or perhaps provide another, consistent mechanism for validating the input of `.plus()` (or its output).

**To Reproduce**
```js
let d = DateTime.local();
let dz = d.setZone("America/Vancouver");

d.plus({seconds: 12345678901234567890});
DateTime {ts: 1.234567890285102e+22, _zone: LocalZone, loc: Locale, invalid: Invalid, weekData: null, …}

dz.plus({seconds: 12345678901234567890});
IANAZone.js:40 Uncaught RangeError: Invalid time value
    at DateTimeFormat.formatToParts (<anonymous>)
    at partsOffset (IANAZone.js:40)
    at IANAZone.offset (IANAZone.js:158)
    at adjustTime (datetime.js:154)
    at DateTime.plus (datetime.js:1379)
    at <anonymous>:1:4
```

**Expected behavior**
Passing a ridiculously high value to an "unzoned" DT object `.plus()` creates an object for which `.isValid === true`, as with `DateTime.local().plus(ridiculous_value)` for example.

**Actual behavior**
If `.setZone(z)` has previously been called, then `.plus(ridiculous_value)` throws the above `RangeError`.

**Environment:**

- OS: Chrome OS 89.0.4389.95 (Official Build) (64-bit)
- Browser: Chrome 89.0.4389.95 (Official Build) (64-bit)
- Luxon version: 1.26.0 (from [here](https://moment.github.io/luxon/demo/global.html))

**Additional context**
I'm using it in the context of adding arbitrary values to .plus() and rejecting those that are out of reasonable range, using `.isValid` and other heuristics in the process of filtering. I can try/catch the error for now but it would be nice to have consistent behavior. [Usage](https://github.com/gumgl/milestones/blob/416eb8e30f73a3aa1549f510c872c99ce583e89e/src/MilestonesList.js#L26).

gumgl:
```js
> d
DateTime {ts: 1616453983301, _zone: LocalZone, loc: Locale, invalid: null, weekData: null, …}
> dz
DateTime {ts: 1616453983301, _zone: IANAZone, loc: Locale, invalid: null, weekData: null, …}
```

[...]

PR #906, opened 2021-03-23
Prevent error from being thrown in IANAZone.offset

gumgl:
To fix bug #905.

I confirmed the fix in the browser (local build):
```js
let d = luxon.DateTime.local();
let dz = d.setZone("America/Vancouver");

d.plus({seconds: 12345678901234567890});
DateTime {ts: 1.2345678902851053e+22, _zone: LocalZone, loc: Locale, invalid: Invalid, weekData: null, …}

dz.plus({seconds: 12345678901234567890});
DateTime {ts: 1.2345678902851053e+22, _zone: IANAZone, loc: Locale, invalid: Invalid, weekData: null, …}
```

### DIFF
--- a/src/zones/IANAZone.js
+++ b/src/zones/IANAZone.js
@@ -152,8 +152,11 @@ export default class IANAZone extends Zone {
 
   /** @override **/
   offset(ts) {
-    const date = new Date(ts),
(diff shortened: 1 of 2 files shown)


louisbuchbinder:
Thank you for your work @gumgl

PR #926, opened 2021-04-30
feat: multiple units support in toRelative

### LINKED ISSUE #579: Feature request: toRelative multiple units
It would be great to specify multiple units for `toRelative()` to receive a more precise result if the date is less than 24 hours.

`toRelative({ units: ['days', 'hours' ]})`

```
> 24 hours

in 2 days
```

---

```
< 24 hours

in 10 hours
```

andreialecu:
This has been requested before in #579, see discussion there for more details.

My use case is that during the first minute, `.toRelative()` will print something like `1 second ago`. I'd like to avoid updating it every second (for performance purposes), and instead I'd like it to print out `0 minutes ago`. This way I can schedule updates once every minute instead of every second.

This PR adds a way to override which units are considered when formatting a relative date.

See the tests I added as well.

Fixes #579.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -360,7 +360,7 @@ function diffRelative(start, end, opts) {
       return format(count, unit);
     }
   }
-  return format(0, opts.units[opts.units.length - 1]);
+  return format(start > end ? -0 : 0, opts.units[opts.units.length - 1]);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


andreialecu:
One thing to note is that this also fixes an issue with rounding:

I added a test for it:
```js
  expect(base.minus({ seconds: 1 }).toRelative({ base, unit: "minutes" })).toBe("0 minutes ago");
```

On `master`, that printed `in 0 minutes` which was wrong.

Issue #927, opened 2021-05-01
Interval.intersection() on abutting intervals returns a 0-length interval instead of null


0x24a537r9:
**Describe the bug**
`Interval.intersection()` on abutting intervals returns a 0-length interval, in spite of https://moment.github.io/luxon/docs/class/src/interval.js~Interval.html#static-method-fromDateTimes saying that `Interval` is "Inclusive of the start but not the end". The docstring for intersection also says explicitly "Returns null if the intersection is empty, meaning, the intervals don't intersect."

**To Reproduce**
```
test('intersection of abutting intervals is null', () => {
  expect(
    Interval.fromISO('2021-01-01T09:00:00Z/2021-01-01T10:00:00Z').intersection(
      Interval.fromISO('2021-01-01T10:00:00Z/2021-01-01T11:00:00Z'),
    ),
  ).toBeNull();
});
```

**Actual vs Expected behavior**
If `Interval` is indeed meant to be inclusive of the start but not the end, then the overlap should be null. Instead I get for the test above:

```
    expect(received).toBeNull()

    Received: {"e": "2021-01-01T02:00:00.000-08:00", "invalid": null, "isLuxonInterval": true, "s": "2021-01-01T02:00:00.000-08:00"}

      1110 |         Interval.fromISO('2021-01-01T10:00:00Z/2021-01-01T11:00:00Z'),
      1111 |       ),
    > 1112 |     ).toBeNull();
           |       ^
      1113 |   });
      1114 |
      1115 |   // tech with null rules
```

It's worth noting that `Interval.overlaps()` works as expected, with the following test passing:

```
test('overlaps of abutting intervals is false', () => {
  expect(
    Interval.fromISO('2021-01-01T09:00:00Z/2021-01-01T10:00:00Z').overlaps(
      Interval.fromISO('2021-01-01T10:00:00Z/2021-01-01T11:00:00Z'),
    ),
  ).toBe(false);
});
```

Conceptually, I think that `Interval.overlaps()` should be perform identically to `Interval.intersection() != null`.

**Desktop (please complete the following information):**

- OS: Mac
- Browser: Node
- Luxon version: `^1.26.0`

PR #928, opened 2021-05-01
#927: Return null when calling intersection() on abutting Intervals

0x24a537r9:
See #927 for the justification.

### Tested
Updated test to intended behavior without updating code and it failed (as intended):
```
 FAIL  test/interval/many.test.js
  ● Interval#intersection returns null for adjacent intervals

    expect(received).toBeNull()

    Received: {"e": "2017-05-25T08:00:00.000-04:00", "invalid": null, "isLuxonInterval": true, "s": "2017-05-25T08:00:00.000-04:00"}

      113 |
      114 | test("Interval#intersection returns null for adjacent intervals", () => {
    > 115 |   expect(todayFrom(5, 8).intersection(todayFrom(8, 10))).toBeNull();
          |                                                          ^
      116 | });
      117 |
      118 | test("Interval#intersection returns invalid for invalid intervals", () => {

      at Object.<anonymous> (test/interval/many.test.js:115:58)
```

Updating interval.js caused the test to begin passing and did not cause any other failures:
```
 PASS  test/interval/many.test.js
```

### DIFF
--- a/src/interval.js
+++ b/src/interval.js
@@ -429,7 +429,7 @@ export default class Interval {
     const s = this.s > other.s ? this.s : other.s,
       e = this.e < other.e ? this.e : other.e;
 
-    if (s > e) {
+    if (s >= e) {
       return null;
     } else {
       return Interval.fromDateTimes(s, e);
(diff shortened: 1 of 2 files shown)


icambron:
Looks good, thanks! Will go into the next release

PR #933, opened 2021-05-10
Update datetime.js

dimacros:
Improve js date creation because it doesn't change correctly when working with timezones.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1759,7 +1759,21 @@ export default class DateTime {
    * @return {Date}
    */
   toJSDate() {
-    return new Date(this.isValid ? this.ts : NaN);
+    if (! this.isValid) {
+        return new Date(NaN);
+    }
+
+    const dateObject = this.toObject();
+
+    return new Date(
+        dateObject.year, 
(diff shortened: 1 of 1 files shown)


icambron:
No, this is incorrect. The JS date represents a specific instant in time and expresses that in the system's zone (for example, the time is `1621364253939 milliseconds epoch` and locally that's  `'Tue May 18 2021 14:58:19 GMT-0400 (Eastern Daylight Time)'`. A Luxon DateTime represents a specific instant in time and expresses it in an *arbitrary* zone. To translate from Luxon to the native Date, the thing we preserve is the instant in time, not how it's expressed. This is really important, because the time is the time, regardless of which zone you express the hours in. This change reverses that, making the thing we preserve the way the time is expressed. But because that's in a different zone, it results in a different underlying time, which would be a disaster for most applications

That doesn't make your use case invalid, just means that's not the expected behavior of `toJSDate()`, and we're not going to change it. To accomplish what I'm surmising you want, do this:

```js
dt.setZone("local", { keepLocalTime: true}).toJSDate();
```

Issue #692, opened 2020-04-15
DateTime.toFormat('ZZZZ') not showing expected output (Again)


philiprabbett:
Previously an issue #499 has raised itself again. To recap:

> According to the [docs](https://moment.github.io/luxon/docs/manual/formatting.html#table-of-tokens), when the `ZZZZ` token is passed in the format string, the abbreviated named offset string should be displayed in the output string.
> However, the timezone offset from GMT is being displayed instead.

Code snippet to test the issue:

```JS
var iso = DateTime.fromISO('2020-01-31T12:18:48+0000', {
    zone : 'Europe/Amsterdam'
});
```

```JS
iso.toFormat('dd MMMM yyyy HH:mm ZZZZ');
```

Expected Output:

> 31 January 2020 13:18 CET

Actual Output:

> 31 January 2020 13:18 GMT+1

```JS
iso.offsetNameShort;
```

Expected Output:

> CET

Actual Output:

> CET

As you can see, we are getting two different results from what should be the same result for the timezone offset name.

Briefly looking into this, this seams to be a locale issue in that the `.offsetNameShort` will use the users / system locale (in my case 'en-GB') while `.toFormat()` is forced to use the locale `en-US' thereby producing two different results.

I've tracked this down to the following place:

https://github.com/moment/luxon/blob/e1f3bb7f7378ea665998079a6763e4bb86f41851/src/impl/locale.js#L290
https://github.com/moment/luxon/blob/e1f3bb7f7378ea665998079a6763e4bb86f41851/src/datetime.js#L1460

To recreate in more detail:

```JS
iso.zone.offsetName(iso.toSeconds(), { format: 'short', locale: 'en-US' })); //=> GMT+1
iso.zone.offsetName(iso.toSeconds(), { format: 'short', locale: 'en-GB' })); //=> CET
```

I'm fine with it being different based on the locale as that's what the browsers are doing by default but the problem is that the same calls via two different methods `.offsetNameShort` and `.toFormat()` results in two different short names meaning `.toFormat()` is not using the same locale as `.offsetNameShort`

philiprabbett:
Oh it looks like this is intentional.

`.offsetNameShort` Defaults to the system's locale if no locale has been specified.
`.toFormat()` Defaults to en-US if no locale has been specified, regardless of the system's locale.

icambron:
Yeah, this is a tricky one. It's pretty unfortunate that en-US returns something different there; in many ways, that's the real issue.

[...]

PR #938, opened 2021-05-17
Clarify toFormat docs

andrey-p:
I'm a dev migrating from Moment.js who just lost a couple of hours to the fact that `toFormat` doesn't default to the user's own locale. I figured I might help others by adding some tweaks to the docs.

My changes are two-fold:

1. Clarify that `format` in Moment.js and `toFormat` in Luxon aren't fully equivalent to each other. In Moment.js, `format` is the preferred way of formatting dates, even for human-readable strings.
2. Add a note to the Intl section of `toFormat` to clarify that the default locale is `en-US`.

Re #2, in the contributing guidelines you say that you prefer gotchas to go in the docstrings, but considering `toFormat` already has a dedicated section for internationalization and [I'm not the first person to run into this](https://github.com/moment/luxon/issues/692), I hope you'll consider making a minor exception.


icambron:
agreed, thanks

Issue #565, opened 2019-08-31
DateTime.toSeconds returns seconds.milliseconds


ThorstenBux:
Hi,

I'm wondering why DateTime.toSeconds() does return a floating-point number xxxxx.yyy where yyy represents the milliseconds. If I wanted to have milliseconds I would have called .toMilliSeconds()

Expected:
DateTime.fromISO("2019-08-31T00:27:39.689Z").toSeconds() // --> 1567211259

Actual:
DateTime.fromISO("2019-08-31T00:27:39.689Z").toSeconds() // --> 1567211259.689

icambron:
The reason people typically want `toSeconds` is that they want to pass the result a UNIX system that expects datetimes expressed as epoch seconds, and I believe those systems generally allow for decimals in the input. Thus, leaving in the milliseconds makes the conversion to that representation lossless.

You can alway call `Math.floor()` or `Math.round()` on it.

ThorstenBux:
@icambron, thanks for clarification.

krukru:
Hey @icambron, just found this issue after spending some considerable time debugging why my app doesn't work, because I would never expect toSeconds() to return seconds.miliseconds. 
I think toSeconds should return exactly what the name suggests, and allow for a different method to return both seconds and milis.
Even better  would be to explicitly  return the result as an object, with properties seconds and miliseconds

vvo:
This was also a surprise to me

carlholloway:
Likewise. I think at the very least the docs should be explicit that a decimal number is returned

toddbluhm:
I just ran into this issue as well. I definitely think it should return just the seconds stripping the millis.

While I totally agree that `toSeconds` is primarily used for Unix systems (that's what I use it for too), the method is called `toSeconds` not `toUnix`. I would be completely happy with this behavior in a `toUnix` method, but `toSeconds` is pretty explicit about what it should return.

I would not change this to return an `object` as that is even less intuitive than the current behavior. `toSeconds` screams number, not object. 

One other possibility is to have an option to enable milliseconds. Something like `includeMillis` in the options object.

Lastly, whether any changes are made, the current behavior should definitely be added/clarified in the docs 😄

joepie91:
[...]

PR #944, opened 2021-05-20
.toSeconds() returns seconds.milliseconds

ebisbe:
As explained on https://github.com/moment/luxon/issues/565 .toSeconds() returns seconds.milliseconds to mimic UNIX system behaviour.


icambron:
Thanks!

PR #946, opened 2021-05-24
Update support matrix

sandstrom:
Node 6 is EOL (https://nodejs.org/en/about/releases/) so I removed it. Also removed other browsers that are very old.

Updated matrix to reflect that Safari now support relative time formatting since 14.4.

### DIFF
--- a/docs/matrix.md
+++ b/docs/matrix.md
@@ -11,13 +11,13 @@ Luxon officially supports the last two versions of the major browsers, with some
 | Chrome                           | >= 71    | ✓     | ✓           | ✓           | ✓                             |
 |                                  | 54-70    | ✓     | ✓           | ✓           | ✗                             |
 | Firefox                          | >= 65    | ✓     | ✓           | ✓           | ✓                             |
-| Edge                             | >= 79    | ✓     | ✓           | ✓           | ✗                             |
+| Edge                             | >= 79    | ✓     | ✓           | ✓           | ✓                             |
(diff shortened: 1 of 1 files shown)


darrinmn9:
@sandstrom just wanted to clarify how you came to this, because maybe i am reading `caniuse` wrong. It seems that recent versions of edge, firefox, chrome, safari and safari on iOS all support INTL relative time. Am i incorrect? or has this been updated since ur PR?

https://caniuse.com/?search=relativetimeformat
Screen Shot 2021-06-09 at 4 30 32 PM

sandstrom:
Yes, all the current ones support it. That's what I tried to express in the table too.

I also listed which versions doesn't support it. But in case that's too much info, we could also just list the minimum version for all browsers.

darrinmn9:
gotcha, it just looks like a few of the `✗`'s in the `Intl relative time formatting` column could be changed to `✓`. (like mobile safari and edge)

sandstrom:
@darrinmn9 You're right, thanks!

sandstrom:
@icambron Let me know what you think!

icambron:
I'll break support on Node 6 in the next minor, yeah. Which will be nice because it simplifies a lot of stuff for Luxon.

icambron:
I ended up being much more aggressive and simply releasing 2.0 with huge changes to this matrix.

sandstrom:
@icambron Awesome! 

Small suggestion here: https://github.com/moment/luxon/pull/969

PR #964, opened 2021-06-29
Fix `install` link

dasa:


dasa:
Fixed in 9f98f827861d2fb0b9b120dc2c0560c3c523a6a3

PR #980, opened 2021-07-09
Docs: Remove outdated note on relative time formatting support

dasa:
Relative time formatting is now also supported in Edge and Safari, as shown in the table and at https://caniuse.com/mdn-javascript_builtins_intl_relativetimeformat


icambron:
Thanks

PR #995, opened 2021-07-22
Remove outdated parsing tokens note

dasa:
Removed:
> That includes all the "macro" formats like "D" for "localized numeric date".

`D` is in the [table](https://moment.github.io/luxon/#/parsing?id=table-of-tokens) and is working, so I assume this is outdated.

`luxon.DateTime.fromFormat("22/07/2021", "D", { locale: "en-GB" }).toISO()` -> "2021-07-22T00:00:00.000-07:00"


icambron (approved):
Yup, thanks

PR #1020, opened 2021-08-27
added fromSeconds, fromMinutes etc. to Duration

mjschurig:
I added fromSeconds, fromMinutes etc. to Duration in the same style as the existing fromMillis method.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -213,6 +213,97 @@ export default class Duration {
   }
 
   /**
+   * Create Duration from a number of seconds.
+   * @param {number} count of seconds
+   * @param {Object} opts - options for parsing
+   * @param {string} [opts.locale='en-US'] - the locale to use
+   * @param {string} opts.numberingSystem - the numbering system to use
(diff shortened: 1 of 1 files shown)


icambron:
Sorry for the slow response, but I'm going to pass on this; they're of too marginal utility. From the contributing guide:

> Luxon shouldn't contain simple conveniences that bloat the library to save callers a couple lines of code. Write those lines in your own code.

mjschurig:
I added this to add functionality similar to .NET timespan, confer to e.g.
https://docs.microsoft.com/de-de/dotnet/api/system.timespan.fromdays?view=net-6.0

This is indeed marginal, as you say, but i found it very convenient as i am
regularly using .NET and C#

btw, writing Duration.FromMillis(days*1000*3600*24*7) instead of
Duration.FromDays(days) is not very nice or readable.

Best regards,
Max

Isaac Cambron ***@***.***> schrieb am Mo., 9. Mai 2022, 01:06:

Issue #973, opened 2021-07-06
IANAZone.isValidZone(undefined) returns true


mscheper:
**Describe the bug**
A call to `IANAZone.isValidZone(zone)`, where `zone` is undefined, returns **`true`**. This is unexpected and undocumented.

**To Reproduce**
Please share a minimal code example that triggers the problem:

```js
const zone = undefined;
console.log(IANAZone.isValidZone(zone)); // true
```

**Actual vs Expected behavior**
I would expect this function to return `false` for an undefined zone.

**Desktop (please complete the following information):**

- OS: Mint 19
- Browser: Chrome 91
- Luxon version: 1.27.0
- Your timezone: America/New_York

socraticDevBlog:
@icambron this open issue seems to have been fixed and ready to be closed, no?  i'm shopping for a bug to fix this morning ;)

icambron:
Sure, thanks

PR #1026, opened 2021-09-09
Protect IANAZone.isValidZone from undefined

benscrane:
This addresses #973 (IANAZone.isValidZone(undefined) returns true).

`Intl.DateTimeFormat("en-US", { timeZone: zone }).format()` will succeed if `zone` is `undefined`.  `null`, and other non-`string` values for timeZone fail.

This adds a check to make sure `zone` is not undefined, as well as a few additional test cases for non-string values.


icambron:
Thanks!

PR #1031, opened 2021-09-21
Fix unclear docs on relative time support

loganvolkers:
The docs seems to indicate that relative time format does not exist, but in fact it does.


icambron:
Thanks!

PR #1035, opened 2021-09-23
Clarify confusing math example

jonathanberger:
The example assumes the variable 'dur' is 1 year but doesn't show that step.


icambron:
Thanks

Issue #1057, opened 2021-10-16
Make friendlyDuration public (`Duration.fromDurationLike`)?


nazarvovk:
There's a neat little function [here](https://github.com/moment/luxon/blob/b3aec6c405806d90dcea593737df77dafe063257/src/duration.js#L848), and I don't see a reason for it to be private.

It converts from `DurationLike` to `Duration`. This won't add any overhead, so maybe making it a public static method `Duration.fromDurationLike` makes sense?

I can make a PR, just say the word :)

PR #1062, opened 2021-10-22
Add fromDurationLike method

nazarvovk:
This PR moves the private `friendlyDuration` function to `Duration.fromDurationLike`.
`DurationLike` is a TS interface defined [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/22d5e137fff20a3da5ba2c4937f10731a100a530/types/luxon/src/duration.d.ts#L60)

This change was suggested in #1057

Basically, no code was added, so bundle size remains the same.

## Why this is useful
This change allows to do things like this:
```ts
type Example = {
  durationProp: Duration
}

const createExample = (duration: DurationLike): Example => {
  return {
    durationProp: Duration.fromDurationLike(duration)
  }
}

const a = createExample(1000)
const b = createExample({ seconds: 5 })

// now we can compare a.durationProp and b.durationProp
```

### DIFF
--- a/test/duration/create.test.js
+++ b/test/duration/create.test.js
@@ -90,3 +90,41 @@ test("Duration.fromObject is valid if providing options only", () => {
   expect(dur.milliseconds).toBe(0);
   expect(dur.isValid).toBe(true);
 });
+
+//------
+// .fromDurationLike()
+//-------
+
+it("Duration.fromDurationLike returns a Duration from millis", () => {
+  expect.assertions(2);
+
+  const result = Duration.fromDurationLike(1000);
+
+  expect(result).toBeInstanceOf(Duration);
+  expect(result).toMatchInlineSnapshot(`"PT1S"`);
+});
+
+it("Duration.fromDurationLike returns a Duration from object", () => {
+  expect.assertions(2);
+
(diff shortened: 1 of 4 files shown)


test/duration/create.test.js
```diff
@@ -90,3 +90,41 @@ test("Duration.fromObject is valid if providing options only", () => {
...
+  expect.assertions(2);
+
+  const result = Duration.fromDurationLike(1000);
+
+  expect(result).toBeInstanceOf(Duration);
+  expect(result).toMatchInlineSnapshot(`"PT1S"`);
+});
+
+it("Duration.fromDurationLike returns a Duration from object", () => {
+  expect.assertions(2);
```
icambron, line 108: You don't need to expect the assertion count. Asserting the count is important if you're doing async where the assertions are in callbacks that might not get called. But here we're not worried that our assertions won't happen

nazarvovk: Yep, fair enough, I just have a lint config in another project that complains if there's no `expect.assertions()` in every test. So it became a bit of habit.
Will remove :)

icambron:
Thanks!

PR #1066, opened 2021-10-29
perf: Use memoized zone for universal zone validation

cotillion:
When formatting to UTC IANAZone.isValidZone will construct a new Intl.DateTimeFormat for each call.

luxon-isvalidzone

### DIFF
--- a/benchmarks/datetime.js
+++ b/benchmarks/datetime.js
@@ -54,6 +54,9 @@ function runDateTimeSuite() {
       .add("DateTime#toLocaleString", () => {
         dt.toLocaleString();
       })
+      .add("DateTime#toLocaleString in utc", () => {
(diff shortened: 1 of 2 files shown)


cotillion:
I added a benhcmark for this issue.

Before:
DateTime#toLocaleString x 184,149 ops/sec ±0.70% (93 runs sampled)
DateTime#toLocaleString in utc x 6,443 ops/sec ±5.20% (86 runs sampled)

After:
DateTime#toLocaleString x 185,117 ops/sec ±0.28% (96 runs sampled)
DateTime#toLocaleString in utc x 86,790 ops/sec ±0.90% (94 runs sampled)

icambron:
Looks good, thanks

PR #1078, opened 2021-11-10
Fix for float comparison issue in shiftTo

bluSCALE4:
Math isn't really my forte, but this is the location of the issue and a possible fix w/ test.


icambron:
I took this fix in 6f86ca38 with a somewhat different test. Thanks for the contribution!

Issue #1076, opened 2021-11-10
fromFormat should prefer the IANA zone if there are multiple zone specifiers


gerob311:
It currently does not seem to be possible to set a time zone on a DateTime when parsing a Java-style timestamp with a time zone.

```
describe('DateTime', () => {
    const formatString = `yyyy-MM-dd'T'HH:mm:ss.SSSZZ'['z']'`
    it('should create from timezoned string', () => {
        const time = DateTime.fromFormat('2021-11-10T13:48:00.000+08:00[Australia/Perth]', formatString, { zone: undefined });
        expect(time.zoneName).toBe('Australia/Perth');
    });
});
```

Expected: Time zone of `time` should be `Australia/Perth`
Actual: Time zone of `time` is my machine's local time zone

I'm a bit confused by this, because the documentation seems to suggest that the opts Object can override a zone specified in the parsed string, which implies to me that it should be possible to specify a zone in the parsed string. 

If I use `fromFormatExplain` instead, the time zone is listed in the `matches` object, but not the `result` object.

```
  {
      input: '2021-11-10T13:48:00.000+08:00[Australia/Perth]',
      tokens: [
        { literal: false, val: 'yyyy' },
        { literal: false, val: '-' },
        { literal: false, val: 'MM' },
        { literal: false, val: '-' },
        { literal: false, val: 'dd' },
        { literal: true, val: 'T' },
        { literal: false, val: 'HH' },
        { literal: false, val: ':' },
        { literal: false, val: 'mm' },
        { literal: false, val: ':' },
        { literal: false, val: 'ss' },
        { literal: false, val: '.' },
        { literal: false, val: 'SSS' },
        { literal: false, val: 'ZZ' },
        { literal: true, val: '[' },
        { literal: false, val: 'z' },
        { literal: true, val: ']' }
      ],
      regex: /^(\d{4})(\-)(\d{2})(\-)(\d{2})(T)(\d{2})(:)(\d{2})(:)(\d{2})(\.)(\d{3})(([+-]\d{1,2})(?::(\d{2}))?)(\[)([a-z_+-/]{1,256}?)(\])$/i,
      rawMatches: [
        '2021-11-10T13:48:00.000+08:00[Australia/Perth]',
        '2021',
        '-',
        '11',
        '-',
        '10',
        'T',
        '13',
        ':',
        '48',
        ':',
        '00',
        '.',
        '000',
        '+08:00',
        '+08',
        '00',
        '[',
        'Australia/Perth',
[...]

PR #1082, opened 2021-11-15
1076: fromFormat should prefer the IANA zone if there are multiple zone specifiers

brendan-gero-humanetix:
This PR is an attempt to address this issue: https://github.com/moment/luxon/issues/1076
I've tried to take the approach of minimising modifications to the public API. I'm not completely confident in some of the changes this has led to, especially in the fromSQL function.


brendan-gero-humanetix:
Could I get some help with the issue with the EasyCLA check? 

- My Github profile has my email address
- I've signed the OpenJS Foundation Corporate CLA as my employer's Initial CLA Manager
- My email address matches the domain that I have added under "Approved List Of Contributors From My Organization"

icambron:
I'm just going to accept the CLA signature as-is. I think you've done it correctly; I'm not sure why the automation isn't accepting it. But it's good by me.

The change itself I have some feedback on. So first, a mea culpa: I had misunderstood your question in the issue thread and answered too quickly. I didn't realize you were positioning it to use the numerical offset to interpret and then converting to the IANA zone. My mistake.

I had in mind just changing the precedence to:

```js
  let zone;
  if (!isUndefined(matches.z)) {
    zone = IANAZone.create(matches.z);
  } else if (!isUndefined(matches.Z)) {
    zone = new FixedOffsetZone(matches.Z);
  } else {
    zone = null;
  }
```

That's it; in other words, I'm suggesting totally ignoring the numerical offset.

Looking at the tests, I think I see why you did the more elaborate change: it allows the parser to resolve ambiguous DST times using the offset. I can see the merits of that.

But the semantics here are too complicated: using `setZone: true`, we interpret using the offset and convert to the IANA zone. I get why that's a convenient way to implement the ambiguous-interpretation, but that's just too messy, and you can see how awkward it sounds in the docstrings. In particular, the `DateTime.fromFormat() converts offset not belonging to time zone` is too strange. That should give 9 hours, not 6, because it should ignore the erroneous 11. I'd rather it throw an error than do unexpected conversions like that. I should have caught that in your first question (again, my mistake).

So here's a proposal to get the best of both worlds:

 1. If one is there but not the other (doesn't matter which), then use that for both interpretation and (if setZone is true) conversion
 2. If both are there, use IANA as the authority for both interpretation and conversion over the numeric offset
 3. If both are there but the time is ambiguous, use the numerical offset to inform which side of the ambiguity is taken

3 sounds hard but really isn't. Here's my version: #1083

brendan-gero-humanetix:
@icambron Sorry I haven't gotten back to you sooner, I've gotten a bit distracted.

I think your PR does make sense. I'd be tempted to go one step further and say that a timestamp with an offset that is completely different to the IANAZone should actually be invalid. However, I don't think I'll be dealing with timestamps like that in my situation, so I'd personally be happy if your PR was accepted as is. 

Should we reject this PR then?

Issue #951, opened 2021-06-02
DateTime toISO pads to 4 instead of 6 digits for negative years


Teedious:
**Describe the bug**
DateTime toISO() function does not pad enough zeros when converting DateTime with negative years to ISO.
Years are padded to **four** digits when they should be padded to **six** as stated in https://github.com/moment/luxon/issues/774#issuecomment-695229272.

**To Reproduce**
`let v = DateTime.fromObject({years:-1}).toISO()`

**Actual vs Expected behavior**
 `v == '-0001-01-01T00:00:00.000+00:53'` [actual]
 `v == '-000001-01-01T00:00:00.000+00:53'` [expected]

**Desktop (please complete the following information):**

- OS: [Windows 10]
- Browser [NodeJS  v14.15.1}
- Luxon version [1.27.0]

This is my first bug report. Please tell me if I should provide anything else.

icambron:
Fixed in #1108

PR #1101, opened 2021-12-25
fix negative year ISO  6 digits padding issue 951

socraticDevBlog:
as requested : negative ISO date years should be 6 digits long

fix broke existing test for ISODate year padding. 

I've adjusted expected values to have six digits negative year instead of 4


socraticDevBlog:
fix issue https://github.com/moment/luxon/issues/951

icambron:
Thanks for this contribution. However, I ended up fixing this in #1108, which refactors all the ISO logic This PR would conflict heavily with that one, so it was easier to fix it there than try to sort out the merge. The fix looks completely different there.

PR #1108, opened 2022-01-02
Continuation of #1097

icambron:
@gpaucot Added some commits:

 1. I really wanted to push the work into `toSQLDate` and `toSQLTime`. So I set up both your fully-inlined one and my refactor and benchmarked them against each other, then tweaked mine until it was within the benchmark's noise level.
 2. Admittedly, my code benefited from some tweaks to the logic, and once I backported those into the inlined code, that code got faster again. But only by a little, so I'm going to live with it.
 3. I also removed some no-longer-needed code and did some other cleanup

These were the logic changes relative to your PR:

 1. Turns out, we're supposed to always use 6-digit years for negative years. That's actually #951. That has a small perf advantage because it's quicker to compare `o.c.year > 9999 || o.c.year < 0` than it is to do `Math.abs(o.c.year) > 9999`. So we'll squash that bug by-the-by.
 2. You had logic for `allowZ`, but that's not actually an option on `toISO`. You always get Z for the UTC zone; the old code hardcoded into the call to the formatter. No real perf difference either way

### DIFF
--- a/benchmarks/datetime.js
+++ b/benchmarks/datetime.js
@@ -24,11 +24,11 @@ function runDateTimeSuite() {
       .add("DateTime.fromSQL", () => {
         DateTime.fromSQL("2016-05-14 10:23:54.2346");
       })
-      .add("DateTime.fromString", () => {
-        DateTime.fromString("1982/05/25 09:10:11.445", "yyyy/MM/dd HH:mm:ss.SSS");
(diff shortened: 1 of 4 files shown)


icambron:
Actually just going to merge this. More perf changes can get tacked on on top of it. Thanks for your help @gpaucot!

PR #1114, opened 2022-01-07
Add whole second unix timestamp function

### LINKED ISSUE #1109: Add toWholeSeconds, toUnixTimestamp or some equivalent that returns an integer
I have no reason to have my unix timestamps in decimal seconds, just using toSeconds would cause a variety of confusing problems that I don't want to have to deal with.

A convenience method that eliminates the above headache

Alternatives suggested are:

- use Math.floor/ceil each time, which adds code bloat and is easy to forget/screw up
- Monkey patch DateTime, which is an anti-pattern and would go outside the TypeScript typings, causing developer confusion and error

**Additional context**
I use moment.unix in loads of spots in my code as my database uses unix seconds in an int type, keeping it that way is all I want. Moment is deprecated, and while I don't have to move to Luxon, I'd like to. But this just seems like a step backwards considering the traditional use of unix timestamps (whole number, in seconds).

jpike88:
fixes #1109

@icambron I used the function name toUnixInteger, but toUnix I guess works too. Just let me know and I'll tweak

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1749,6 +1749,14 @@ export default class DateTime {
   toSeconds() {
     return this.isValid ? this.ts / 1000 : NaN;
   }
+  
+  /**
+   * Returns the epoch seconds (as a whole number) of this DateTime.
+   * @return {number}
+   */
+  toUnixInteger() {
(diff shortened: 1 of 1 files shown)


icambron:
Thanks!

PR #1116, opened 2022-01-08
Add luxon to console in /api-docs

nkalupahana:
Having `luxon` accessible in the console of https://moment.github.io/luxon/ is super useful to me -- however, it isn't available on /api-docs, where it is arguably even more useful. Many times, I've opened the console to try something in the API docs out, just to find that `luxon` isn't there. (Most of the time, I end up switching tabs back to the homepage to run it in the homepage console.) This PR injects `luxon.js` into /api-docs so it's accessible there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "build-global": "babel-node tasks/buildGlobal.js",
     "jest": "jest",
     "test": "jest --coverage",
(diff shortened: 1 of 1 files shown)


nkalupahana:
@icambron or anyone else, I'd love to get a review on this whenever possible :)

Issue #1050, opened 2021-10-11
Duration.diff() returns -0


basnielander-winvision:
The following method
```
public static getTimeDifferenceInMinutes(dateTimeA: Date, dateTimeB: Date): number {
    const duration = DateTime.fromJSDate(dateTimeA).diff(DateTime.fromJSDate(dateTimeB), ['minutes', 'milliseconds']).toObject();
    
    return duration?.minutes ?? 0;
}
```

returns  
![image](https://user-images.githubusercontent.com/70693955/136766890-13397303-9616-4050-8090-c276220964ed.png)

when running the following test:
```
it('2 Date objects with a time of Date A that differs 59 seconds and some milliseconds with Date B, should return 0', () => {
    // arrange
    const dateA = new Date("2021-10-01T23:04:02.007");
    const dateB = new Date("2021-10-01T23:05:01.087");

    // act
    const actual = getTimeDifferenceInMinutes(dateA, dateB);

    // assert
    expect(actual).toBe(0);
});
```

I would expect duration.minutes to be 0.

- OS: Windows 10
- Luxon version: v2.0.2
- TypeScript: v3.9.10
- Timezone [e.g. "Europe/Amsterdam"]

sdotson:
Opened a pr...

PR #1122, opened 2022-01-20
Update Duration.negate() to not negate zeros

sdotson:
Fixes https://github.com/moment/luxon/issues/1050

Appears when diffing two dates and the second date is larger. We negate all units, leading to confusing values such as `-0` for units with no difference.


icambron:
Thanks!

PR #1144, opened 2022-02-23
Fix bad link

elliott-king:
See current location: https://moment.github.io/luxon/api-docs/index.html#durationtohuman

Link is attempting to go back to internal docs, and the page does not exist. 

I didn't find an example of a shorthand link going to Mozilla, so I removed shorthand link and added Mozilla link in `@see`.


icambron:
Thanks!

PR #1158, opened 2022-03-17
removing duplicate comment in datetime.js

nick-krantz:
I came across while learning about DateTime, I figured it was worth a small PR!


nick-krantz:
I closed this to make an edit to the commit

PR #1173, opened 2022-04-01
feat: add week formatting token "w" for duration

### LINKED ISSUE #1164: `Duration#toFormat` should support a formatting token for `week`
I'm surprised that `Duration#toFormat` doesn't support a token for `week` (7 days). (e.g. `w`)

Is there a reason this unit token was omitted for Duration specifically?

Current release tag source ref: https://github.com/moment/luxon/blob/2.3.1/src/duration.js#L397-L426

jsejcksn:
Resolves #1164

(Working with casual conversion accuracy was quite confusing when writing the test. 🤯)

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -401,6 +401,7 @@ export default class Duration {
    * * `m` for minutes
    * * `h` for hours
    * * `d` for days
+   * * `w` for weeks
    * * `M` for months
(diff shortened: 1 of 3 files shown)


icambron:
Looks good. This will go out in the next release, which is soon

PR #1181, opened 2022-04-13
style(demo): add background on hover

depyronick:
adds background color to rows on demo to make it simple to match example to it's result.


icambron:
Sure, thanks!

Issue #1165, opened 2022-03-27
Duration.diff returns negative number


joshua0308:
**Describe the bug**
I get a negative number for a diff although the end date is after the start date.

**To Reproduce**
```js
const start = DateTime.fromISO("2022-05-05T23:00", { zone: "UTC" });
const end = DateTime.fromISO("2022-05-10T00:00", { zone: "Europe/Madrid" }); // 2022-05-09T22:00Z
const diff = end.diff(start, ['days', 'hours', 'minutes']).toObject();

// diff === { days: 4, hours: -1, minutes: 0 }
```

**Actual vs Expected behavior**
I expect the diff to equal `{ days: 3, hours: 23, minutes: 0}` instead of `{ days: 4, hours: -1, minutes: 0 }`. It seems unintuitive to me that I have to think about adding and then subtracting to get to the diff.

**Desktop (please complete the following information):**

- OS: Mac OS
- Browser Chrome 99
- Luxon version 2.3.1
- Your timezone "America/New_York"

JoaoAugustoPansani:
May I try to fix this?

icambron:
@JoaoAugustoPansani sure

JoaoAugustoPansani:
Hi @icambron , I don't have access to the repo, so I'll propose the solution: 

On the file `./scripts/src/impl/diff.js` there's a function `dayDiff`:

```
function dayDiff(earlier, later) {
  const utcDayStart = (dt) => dt.toUTC(0, { keepLocalTime: true }).startOf("day").valueOf(),
    ms = utcDayStart(later) - utcDayStart(earlier);
  return Math.floor(Duration.fromMillis(ms).as("days"));
}
```

What I did for solving this issue, was delete the second argument inside `.toUTC`:

```
function dayDiff(earlier, later) {
  const utcDayStart = (dt) => dt.toUTC(0).startOf("day").valueOf(),
    ms = utcDayStart(later) - utcDayStart(earlier);
  return Math.floor(Duration.fromMillis(ms).as("days"));
}
```

Then I wrote a test inside `./scripts/test/datetime/diff.test.js`: 

```
test("DateTime#diff works when date has a zone object", () => {
  const start = DateTime.fromISO("2022-05-05T23:00", { zone: "UTC" });
  const end = DateTime.fromISO("2022-05-10T00:00", { zone: "Europe/Madrid" });

  const diff = end.diff(start, ["days", "hours", "minutes"]).toObject();
  expect(diff.days).toBe(3);
[...]

PR #1186, opened 2022-04-21
#1165: Fixing and Testing "Duration.diff returns negative number"

JoaoAugustoPansani:
On the file `./scripts/src/impl/diff.js` there's a function `dayDiff`:

```
function dayDiff(earlier, later) {
  const utcDayStart = (dt) => dt.toUTC(0, { keepLocalTime: true }).startOf("day").valueOf(),
    ms = utcDayStart(later) - utcDayStart(earlier);
  return Math.floor(Duration.fromMillis(ms).as("days"));
}
```

What I did for solving this issue, was delete the second argument inside `.toUTC`:

```
function dayDiff(earlier, later) {
  const utcDayStart = (dt) => dt.toUTC(0).startOf("day").valueOf(),
    ms = utcDayStart(later) - utcDayStart(earlier);
  return Math.floor(Duration.fromMillis(ms).as("days"));
}
```

Then I wrote a test inside `./scripts/test/datetime/diff.test.js`:

```
test("DateTime#diff works when date has a zone object", () => {
  const start = DateTime.fromISO("2022-05-05T23:00", { zone: "UTC" });
  const end = DateTime.fromISO("2022-05-10T00:00", { zone: "Europe/Madrid" });

  const diff = end.diff(start, ["days", "hours", "minutes"]).toObject();
  expect(diff.days).toBe(3);
  expect(diff.hours).toBe(23);
  expect(diff.minutes).toBe(0);
});
```


icambron:
Sorry for the slow response. I don't think this fix is correct, though. That second argument ensures that diff compares local times, not absolute ones. You can see the problem by using these dates:

```

omasback:
Sorry I've only briefly looked at this but why not just convert both datetimes `toMillis()` and subtract? Then convert millis to Duration and shift back to requested units?

Issue #1169, opened 2022-03-29
Parsing error in Vietnamese


mwath:
**Describe the bug**
An error is thrown while trying to parse a Vietnamese month.
The issue could in fact appear for any locale that has spaces in the month. (i.e. `tháng 3` is `mars` in Vietnamese)

**To Reproduce**
```JavaScript
luxon.DateTime.fromFormat("tháng 3 2022", "MMMM yyyy", {locale: "vi-VN"})
```

**Additional context**
From what I've understood while debugging, Luxon generates the list of all months and creates a regex based on it to parse the date. However, Luxon [replaces all spaces](https://github.com/moment/luxon/blob/master/src/impl/tokenParser.js#L19) in the regex by a group that matches a regular space or a non-breakable space. It introduces more capturing groups than expected and crashes Luxon.

In English, with this code: `DateTime.fromFormat("March 2022", "MMMM yyyy")`
Luxon will generate the following regex: `/^(January|February|March|April|May|June|July|August|September|October|November|December)(\ )(\d{4})$/i` which has 3 capturing groups.

However, in Vietnamese, with the code provided in To Reproduce, Luxon will produce the following regex:
`/^(tháng( | )1|tháng( | )2|tháng( | )3|tháng( | )4|tháng( | )5|tháng( | )6|tháng( | )7|tháng( | )8|tháng( | )9|tháng( | )10|tháng( | )11|tháng( | )12)(\ )(\d{4})$/i` which has obviously more than 3 capturing groups.

I think that changing [`spaceOrNBSP`](https://github.com/moment/luxon/blob/master/src/impl/tokenParser.js#L16) to a non-capturing group would fix the issue, however I do not know if there will be side effects.

icambron:
> I think that changing [spaceOrNBSP](https://github.com/moment/luxon/blob/master/src/impl/tokenParser.js#L16) to a non-capturing group would fix the issue, however I do not know if there will be side effects.

This sounds like a reasonable fix. If it fixes the issue and passes existing tests, I'd take that patch

mwath:
fixed in #1194

PR #1194, opened 2022-05-03
fix: change NBSP regex to a non-matching group #1169

mwath:
Some languages can have spaces or non-breakable spaces in the month definition.
This lead to creating a regex with more capturing groups than it should.

For example, `tháng 3` is `mars` in Vietnamese and contains a space. (see #1169)
And if you try to parse this month using: `DateTime.fromFormat("tháng 3 2022", "MMMM yyyy", {locale: "vi-VN"})`
It will create the following regex: `/^(tháng( | )1|tháng( | )...)(\d{4})$/i`
Which has a capturing group for each space.

After this commit, it will create the following regex instead: `/^(tháng[  ]1|tháng[  ]...)(\d{4})$/i`
So it does not create additional capturing groups.

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -13,7 +13,7 @@ function intUnit(regex, post = (i) => i) {
 }
 
 const NBSP = String.fromCharCode(160);
-const spaceOrNBSP = `( |${NBSP})`;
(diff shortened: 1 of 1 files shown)


icambron:
Thanks!

PR #1199, opened 2022-05-07
fix bug 908 isInDST() incorrect

socraticDevBlog:
this fix the bug for DateTime.isInDST() returns False when it should return True.

 specifically for the year 1974 in USA.

from January 1974 to end of October 1974, USA was in DST

this is a very special problem that is not really fixable with an elegant algorithm.

this is my proposition (far from perfect). let me know what you think about it.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -42,6 +42,7 @@ import {
   InvalidDateTimeError,
 } from "./errors.js";
 import Invalid from "./impl/invalid.js";
+import IANAZone from "./zones/IANAZone.js";
 
 const INVALID = "Invalid DateTime";
 const MAX_DATE = 8.64e15;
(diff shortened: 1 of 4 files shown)


hawkett:
An alternative could be to just check explicitly against Jan 1 in all circumstances (rather than solving for just USA in 1974) - e.g. just add `day: 1`:

```
return (
    this.offset > this.set({ month: 1, day: 1 }).offset || this.offset > this.set({ month: 5 }).offset
);
```

The existing method checks a random day in January, while this approach just forces that day to be consistently the 1st of January. This has the advantage of capturing other possible occurrences, past and future.

socraticDevBlog:
omg!  thanks @hawkett

icambron:
Thanks. Yeah, the amended fix is much better. If we have to check more dates, we can do that too.

Issue #1182, opened 2022-04-14
setZone('system') is using the Timezone in Settings.defaultZone, not the browser Timezone


benjaminpick:
**Describe the bug**
When using an explicit value in Settings.defaultZone (e.g. 'Asia/Kolkata'), the method `setZone('system')` is not changing the timezone - because 'system' appears to mean "whatever is set in Settings.defaultZone, or, if this is set to 'system', then use the Browser Timezone."

**To Reproduce**
https://codesandbox.io/s/modest-haze-s766q6?file=/src/index.js
```js
const LOCAL_TIMEZONE = "Asia/Kolkata";

Settings.defaultZone = LOCAL_TIMEZONE;
const dt = DateTime.now(); // 2022-04-14T21:56:54.028+05:30 (LOCAL_TIMEZONE)
const zoned = dt.setZone("system", { keepLocalTime: true }); // 2022-04-14T21:56:54.028+05:30 (actual: LOCAL_TIMEZONE, expected: Browser Timezone)

Settings.defaultZone = "system";
const zoned2 = dt.setZone("system", { keepLocalTime: true }); // 2022-04-14T21:56:54.028+02:00 (Browser Timezone)
```

**Actual vs Expected behavior**
I would have expected `zoned` to be in the same timezone as `zoned`. As far as I see it, there is no other possibility to access the "system" timezone without changing the defaultZone. 

I find this confusing because the documentation says: 
`By default, Luxon creates DateTimes in the system's zone. However, you can override this behavior globally:`
Also, the keyword 'system' means something else in `setZone('system')` and in Settings.defaultZone

For backwards compatibility, we probably can't change the meaning of the 'system' timezone now. So I suggest to add a new keyword 'browser':

```js
Settings.defaultZone = LOCAL_TIMEZONE;
const dt = DateTime.now(); // 2022-04-14T21:56:54.028+05:30 (LOCAL_TIMEZONE)
const zoned = dt.setZone("browser", { keepLocalTime: true }); // should be 2022-04-14T21:56:54.028+02:00 (Browser Timezone)
Settings.defaultZone = 'browser'; // should be an alias of 'system'
```
and clean up the wording in the documentation accordingly.

**Use-Case**
In my case, all datetimes are in a certain defaultZone (of a local event), because the times need to be shown in local time. However, a datepicker library expects the dates to be in browser time, so I need to create a "zoned" Date object via 'keepLocalTime'.

icambron:
Yeah, I agree:

[...]

PR #1206, opened 2022-05-13
Add condition for system setZone

davkub:
**What is this about:**
The original issue that led to creating this PR was `setZone('system')` is using the Timezone in `Settings.defaultZone`, not the browser Timezone (https://github.com/moment/luxon/issues/1182), the name is pretty self-explanatory.

**Changes in this PR:**

- added a condition for `system` to return `SystemZone.instance` along with `local`
- added a condition for `defaultZone` to accept also `default` parameter
- Fixed deprecated tests that took local zone as default while setting non-system one, as per "local" is a legacy name for "system"
- Add test for `setZone()` using the `system` zone

### DIFF
--- a/src/impl/zoneUtil.js
+++ b/src/impl/zoneUtil.js
@@ -8,6 +8,7 @@ import FixedOffsetZone from "../zones/fixedOffsetZone.js";
 import InvalidZone from "../zones/invalidZone.js";
 
 import { isUndefined, isString, isNumber } from "./util.js";
+import SystemZone from "../zones/systemZone.js";
 
 export function normalizeZone(input, defaultZone) {
   let offset;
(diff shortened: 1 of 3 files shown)


src/impl/zoneUtil.js
```diff
@@ -17,7 +18,8 @@ export function normalizeZone(input, defaultZone) {
     return input;
   } else if (isString(input)) {
     const lowered = input.toLowerCase();
-    if (lowered === "local" || lowered === "system") return defaultZone;
+    if (lowered === "local" || lowered === "default") return defaultZone;
```
icambron, line 21: "local" is a legacy name for "system", so local should have followed system over to the new condition

icambron:
Try running tests with `scripts/test`, which set some env vars

src/impl/zoneUtil.js
davkub: @icambron This has been refactored, local condition moved to the system one.

davkub:
> Try running tests with `scripts/test`, which set some env vars
@icambron 
I have updated "deprecated" tests that used setZone('local') as a default one, currently local refers to the system zone as you described in your comment. Also added 1 more setZone test for system zone option.

icambron:
Ok, this is ready to go. I just need to decide if I can get away with releasing this as Luxon 2.5 or if it needs to be called 3.0

davkub:
> Ok, this is ready to go. I just need to decide if I can get away with releasing this as Luxon 2.5 or if it needs to be called 3.0

Hello, have you decided which version are we going with?

icambron:
@davkub I'm releasing this in 3.0 today

PR #1211, opened 2022-05-19
Update why.md

mchusma:
I was a bit confused by this, and thought that Safari supported Luxon. But it appears that it now works starting in Safari 14, released in Sep 2020. So updating these docs may help others avoid confusion.

Thanks so much for your work on this library!


bariod:
someone else do it lmao

icambron:
Thanks!

Issue #880, opened 2021-02-23
Duration doesn't parse its own ISO format


thibaultdalban:
**Describe the bug**
`Duration.fromISO` doesn't handle decimal duration.

**To Reproduce**
`luxon.Duration.fromISO(luxon.Duration.fromObject({hours:0.5}).toISO())`

**Actual vs Expected behavior**
Actual
`Object { reason: "unparsable", explanation: "the input \"PT0.5H\" can't be parsed as ISO 8601"`
Expected
A valid Duration object

**Desktop (please complete the following information):**
Latest version.

**Additional context**
The [regex](https://github.com/moment/luxon/blob/master/src/impl/regexParser.js#L127) only handles fraction second (introduced by 5a6256978b28f8a010207b58f12ec178cc71b5d9)

thomas-darling:
Yep, this is indeed broken.

Looks like support for fractional digits was added in https://github.com/moment/luxon/commit/e5a66144b2986237e77206520eac5c4b8319663d, but the regex restricts the number of fractional digits to 9, with the exception of fractional seconds, which is restricted to 20. However, the `toISO` method outputs up to 16 fractional digits, which means the output of `toISO` may not be parsable by `fromISO`.

I've opened a pull request here https://github.com/moment/luxon/pull/1213, which increases the number of allowed digits to 20 for all units.

@icambron Do you think we could get this one merged and released in the near future?
It would be very much appreciated 🙂

PR #1213, opened 2022-05-27
Increase number of allowed digits when parsing ISO duration

thomas-darling:
Fixes https://github.com/moment/luxon/issues/880.

The regex for parsing an ISO duration string restricts the number of fractional digits to 9, with the exception of fractional seconds, which is restricted to 20. However, the `toISO` method outputs up to 16 fractional digits, which means the output of `toISO` may not be parsable by `fromISO`.

This fix increases the number of allowed digits to 20 for all units.


icambron:
Thanks!

Issue #252, opened 2018-05-02
Not working with ESM on Node 10


damianobarbati:
Hey, any chance to have luxon work with Node ESM modules?

```
import { DateTime } from 'luxon';
           ^^^^^^^^
SyntaxError: The requested module 'luxon' does not provide an export named 'DateTime'
at ModuleJob._instantiate (internal/modules/esm/module_job.js:89:21)
```

icambron:
I've never used ESM, so I am not sure. Pretty hard to imagine why it thinks that. E.g. see the end of http://moment.github.io/luxon/node/luxon.js

I'd take a fix to the toolchain that helped here, but I have no idea how to help myself

ericmorand:
Aren't ESM modules an experimental feature on node 10?

damianobarbati:
Yes, and soon moving to harmony flag. Let's be prepared! :)

NB: I've been using them for weeks by now and they work seamlessly (only pain in the ass is you can't import .json as you used to require .json ...).

ericmorand:
>  (only pain in the ass is you can't import .json as you used to require .json ...)

That's actually a very good thing if you ask me. JSON files are not modules in any way and this was leading to a lot of potential problems when you were requiring modules in the blind.

Anyway, yes, ESM is just around the corner. Why is it not working? Isn't doing this enough to support ESM?


`export { DateTime, Duration, Interval, Info, Zone, FixedOffsetZone, IANAZone, LocalZone, Settings };
`

https://github.com/moment/luxon/blob/19b6c26063498806266ab8b00bfae4d6754e7569/src/luxon.js#L11

ericmorand:
Oh yes, good one.

icambron:
Yes, we link to the built one, where the relevant code is:

```js
exports.DateTime = DateTime;
exports.Duration = Duration;
exports.Interval = Interval;
exports.Info = Info;
exports.Zone = Zone;
exports.FixedOffsetZone = FixedOffsetZone;
exports.IANAZone = IANAZone;
exports.LocalZone = LocalZone;
exports.Settings = Settings;
```

damianobarbati:
Maybe .mjs extension is required?

icambron:
Well, even if that's true, I can't just export that or it will break versions of Node that don't support it. What we need is something like:

1. Make the build produce a separate file
 2. Point to it with a different key in package.json that Node will know to import from

[...]

PR #1218, opened 2022-06-09
Add support for ESM-style Node imports

voraciousdev:
Importing `luxon` in a modern Node ESM project with `"type": "module"` causes the following error. 

```
Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
```

Aside from changing the build configuration of this project, these changes are an unobtrusive way to support ESM-style Node projects. This PR will possibly close https://github.com/moment/luxon/issues/252 too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,12 @@
     "immutable"
   ],
   "repository": "https://github.com/moment/luxon",
+  "exports": {
+    ".": {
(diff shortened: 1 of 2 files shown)


icambron:
Ah, this has been requested a few times but I didn't know how to do it.

PR #1220, opened 2022-06-15
Duration support custom matrix

JesusTheHun:
Duration for businesses days are not calendar durations. Meaning that a day is 7 or 8 hours of work and a week probably 5 days.
So if I say a task is going to take 2 days, it means 16 hours of work.

In order to solve that I've added the support for custom matrix for `Duration`. Doing so the `shiftTo` method can transform units as we want it to.

There is a problem is with `DateTime.plus()`. I have noticed it shifts everything to milliseconds to move the clock forward. While I understand why, it raises an issue because it doesn't account for the leap in time between days for exemple. 

`2 days means 14 hours means 840 minutes means 50400 seconds means 50,400,000 millis` according to the duration matrix.

However if you add 2 days to `2022-06-15 05:00` you expect it to move forwards to `2022-06-17 05:00` and not `2022-06-15 19:00`. 

This behaviour, of course falls into a plugin's responsibility.

In order to be able to create such plugin (or extend the existing ones like https://github.com/amaidah/luxon-business-days) I would need the `Duration` to be able to accept a custom matrix. 

Hence this PR.
It's tiny so it doesn't bloat the lib, but it opens more room for extension by plugins.

What do you think ?

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -173,6 +173,11 @@ export default class Duration {
    * @private
    */
   constructor(config) {
+    if (config.conversionAccuracy && config.matrix)
+      throw new Error(
+        "You cannot use a custom matrix along with the `conversionAccuracy` config property"
+      );
+
     const accurate = config.conversionAccuracy === "longterm" || false;
     let matrix = accurate ? accurateMatrix : casualMatrix;
 
--- a/src/duration.js
+++ b/src/duration.js
@@ -120,6 +120,7 @@ function clone(dur, alts, clear = false) {
     values: clear ? alts.values : { ...dur.values, ...(alts.values || {}) },
     loc: dur.loc.clone(alts.loc),
     conversionAccuracy: alts.conversionAccuracy || dur.conversionAccuracy,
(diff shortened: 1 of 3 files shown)


src/duration.js
```diff
@@ -172,7 +173,18 @@ export default class Duration {
    * @private
    */
   constructor(config) {
+    if (config.conversionAccuracy && config.matrix)
+      throw new Error(
```
JesusTheHun, line 177: Not sure if it's ok for you to throw error, but since I couldn't drop the `config.conversionAccuracy` config for obvious reasons, I decided to add a safety guard.

icambron: It's ok to throw an error here, but the errors need to be typed. See https://github.com/moment/luxon/blob/master/src/errors.js

icambron (changes requested):
This looks good and I'll take it if you rework the error

JesusTheHun:
> This looks good and I'll take it if you rework the error

It's done 👌

icambron:
@JesusTheHun thanks

PR #1242, opened 2022-07-14
Fix possible documentation typo

jayremias:


icambron:
I don't follow: `avril` is French for April. The documentation is showing that if you set the locale to `fr`, you get a French string.

PR #1264, opened 2022-08-10
zones.md: assign to defaultZoneName

xathis:
`defaultZone` is readonly.
Current example leads to "Cannot assign to 'defaultZone' because it is a read-only property.ts (2540)". 
Change by setting `defaultZoneName` instead.


icambron:
Thanks!

PR #1271, opened 2022-08-26
Update parsing docs to show syntax for literals

Lull3rSkat3r:
## Why
I could not find a location within the docs where it explains the escape/literal syntax for `fromFormat` when literal values are included in the parsed string (e.g. `T` in `2022-03-15T11:11:11.111Z`). I was able to figure it out by looking at the tests in [`tokenParse.test.js`](https://github.com/moment/luxon/blob/master/test/datetime/tokenParse.test.js#L569), so thought I would bubble it up into the docs for other users.

## What
Added a row to the token table explaining `'` is the token for starting and ending a literal. Albeit this a little weird as the literals aren't technically tokens, but `'` technically is. I can move it to be a part of the paragraph if the core maintainers prefer this. Just let me know!


icambron:
Sure, thanks

Issue #1175, opened 2022-04-03
Many "404 File not found" errors for links in API docs


h-h-h-h:
**Describe the bug**
In the API docs under https://moment.github.io/luxon/api-docs/index.html, there are many invalid links. This isn't exhaustive; it's just an example.

**To Reproduce**
- Visit https://moment.github.io/luxon/api-docs/index.html#datetime
- Click on the first link with text "DateTime#local". The link is: https://moment.github.io/luxon/api-docs/DateTime#local
- The following page says: "404 File not found..."

**Actual vs Expected behavior**
- The correct link would be: https://moment.github.io/luxon/api-docs/index.html#datetimelocal

**Desktop (please complete the following information):**

- OS: Windows
- Browser: Firefox
- Luxon version: N/A
- Your timezone: N/A

**Additional context**
&mdash;

icambron:
Yeah, a lot of these broke when we changed doc systems, and I just haven't tracked them all down

sam-s4s:
Was just about to post this issue. I'm in the process of switching over from moment to luxon, and I keep clicking on all the dead links :P

I'm happy to help update links if someone points me in the right direction?

icambron:
@sam-s4s The doc files are here: https://github.com/moment/luxon/tree/master/docs

A good link looks like this:

```
[time zones documentation](zones.md?id=math-across-dsts)
```

which works both in Github and the rendered site

sam-s4s:
Hmmm, this seems to have everything except the api docs?
https://moment.github.io/luxon/api-docs/index.html
I'm not sure where to find that.

elliott-king:
It looks like all non-sidebar links to _static members_ are not working.

Example of **failing** links (static Datetime members within docs page):
<img width="695" alt="Screen Shot 2022-07-08 at 9 36 13 AM" src="https://user-images.githubusercontent.com/8884571/178003084-0b60e6eb-b569-4187-8fe4-530fd6e4df28.png">


Example of **failing** links (static Duration members within docs page):
<img width="624" alt="Screen Shot 2022-07-08 at 9 36 35 AM" src="https://user-images.githubusercontent.com/8884571/178003247-30c32fa6-8fe7-4e0a-b912-0f82425eee66.png">


Example of **successful** links (static Duration members in sidebar):
[...]

PR #1272, opened 2022-08-26
Fixed JSDocs for static members

huilibaba:
Fixed the JSDocs for static members to fix the links in API docs raised in #1175


icambron:
Thanks

PR #1280, opened 2022-09-03
Add lastMoment getter to Interval

ties-s:
Convenient way to get the last moment that is actually included in the interval. Makes it easier to properly use the fact that intervals are half open.

My use case: displaying end date of a monthly subscription.

I'm not sure `lastMoment` is the best way to name this, but it was the first thing that I thought of.

### DIFF
--- a/src/interval.js
+++ b/src/interval.js
@@ -195,10 +195,10 @@ export default class Interval {
   }
   
   /**
-   * Returns the last moment included in the interval (since end is not part of the interval)
+   * Returns the last DateTime included in the interval (since end is not part of the interval)
    * @type {DateTime}
    */
-  get lastMoment() {
+  get lastDateTime() {
     return this.isValid ? this.e.minus(1) : null;
   }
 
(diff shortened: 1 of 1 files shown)


icambron:
@ties-s So first, sorry for taking _so_ long to get to this.

I'm open to this but a couple of notes:

1. I think `lastMillisecond()` is a better name - the term "moment" is not defined in Luxon's documentation or API
2. It needs test, as simple as those tests would be.

ties-s:
@icambron thank you for your response. 

I'll look into the tests. 

For the name, `lastMillisecond` might suggest a numeric return value. Some other options I thought of, but are also not great:
- `inclusiveEnd`
- `endInclusive`
- `lastInstant`

icambron:
@ties-s sorry a second time for the slow review here. I agree with your concern that `lastMillisecond` sounds numeric. How about `lastDateTime`

ties-s:
@icambron That sounds good, updated the PR to reflect that new function name

diesieben07:
@ties-s Looks good to me! Once we have 3.4.2 out with a fix for the current `Duration` issues, this can be merged.

Issue #1262, opened 2022-08-07
Missing 'at' in FULL and HUGE examples for en_US


dmalan:
As browsers upgrade to ICU 71, dates and times in `en_US` are combined with `at`, per https://github.com/unicode-org/cldr-json/blob/8d565de3afecd4dc3f82b3bdbc3336e2d588a350/cldr-json/cldr-dates-modern/main/en/ca-gregorian.json#L346-L347, as noted in https://github.com/tc39/ecma402/issues/665. 

E.g., on the latest versions of Chrome and Firefox, each of the below now output an `at`:

```js
> luxon.DateTime.now().toLocaleString(luxon.DateTime.DATETIME_FULL)
'July 31, 2022 at 9:50 AM EDT'
> luxon.DateTime.now().toLocaleString(luxon.DateTime. DATETIME_HUGE)
'Sunday, July 31, 2022 at 9:50 AM Eastern Daylight Time'
> luxon.DateTime.now().toLocaleString(luxon.DateTime. DATETIME_FULL_WITH_SECONDS)
'July 31, 2022 at 9:50:00 AM EDT'
> luxon.DateTime.now().toLocaleString(luxon.DateTime. DATETIME_HUGE_WITH_SECONDS)
'Sunday, July 31, 2022 at 9:50:00 AM Eastern Daylight Time'
```

Docs currently omit the `at`, https://github.com/moment/luxon/blob/master/docs/formatting.md#presets.

PR at https://github.com/moment/luxon/pull/1256.

PR #1282, opened 2022-09-10
Update text example of DATETIME_FULL formatting

tylercrumpton:
In #1262, the formatting docs were updated to include examples of the new formatting for `DATETIME_FULL` introduced in ICU 71; this PR updates another one of these examples included further up the page for consistency.


icambron:
thanks

PR #1292, opened 2022-09-26
STAR-41664: Disambiguate method description

robkoenig:


icambron:
Yeah, I agree with @dobon - we can't add this to every method. Think of "it doesn't mutate" as a universal feature of Luxon.

Issue #1233, opened 2022-07-08
`Duration#normalize` has trouble with negative milliseconds


elliott-king:
**Describe the bug**
Fails to normalize a Duration with negative milliseconds.

**To Reproduce**
I was able to repro this in the playground in the docs:
```javascript
> luxon.VERSION
'2.4.0'

> let d = luxon.Duration.fromObject({seconds: -44, milliseconds: -100, minutes: 1})

> d.normalize()
Duration {values: {…}, loc: Locale, conversionAccuracy: 'casual', invalid: null, matrix: {…}, …}
...
values: {seconds: 16, milliseconds: -100, minutes: 0}
```

I would expect the above to be {seconds: 15, milliseconds: 900, minutes: 0}

Out of curiosity, I tried the initializing the "right" duration, and it has no issues normalizing:

```javascript
> let m = luxon.Duration.fromObject({seconds: 15, milliseconds: 900})

> m.normalize()
Duration {values: {…}, loc: Locale, conversionAccuracy: 'casual', invalid: null, matrix: {…}, …}
...
values: {seconds: 15, milliseconds: 900}
```

**Actual vs Expected behavior**
Since the overall duration is greater than zero, I expect all normalized fields to be greater than zero.

**Desktop (please complete the following information):**

- OS: MacOS 11.6.5
- Browser: Chrome Version 102.0.5005.115 (Official Build) (x86_64)
- Luxon version: 2.4.0 (used your playground)
- Your timezone "America/New_York"

**Additional context**
This library is excellent, by the way!

elliott-king:
Possibly unrelated: I would also expect `normalize` to change 66 seconds into a combination of seconds & minutes:
```javascript
> luxon.Duration.fromObject({days: 777}).normalize()
...
values: {days: 777}

> luxon.Duration.fromObject({milliseconds: 15900}).normalize()
...
values: {milliseconds: 15900}
```

icambron:
Yeah, that could definitely use some improvements

CristiMacovei:
@icambron Can I solve this issue?

ZainGulbaz:
@icambron Can you please assign this to me? Thanks.

thomassth:
Hi @icambron , would you mind checking out #1467 ? I've included a test that'd fail in current builds, but passing after PR.

PR #1296, opened 2022-10-04
fixed issue #1233

CristiMacovei:
Made a couple of changes to the code to solve issue #1233

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -125,20 +125,19 @@ function clone(dur, alts, clear = false) {
   return new Duration(conf);
 }
 
-function antiTrunc(n) {
-  return n < 0 ? Math.floor(n) : Math.ceil(n);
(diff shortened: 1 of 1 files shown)


icambron:
Thanks. This will need tests though

CristiMacovei:
It did pass the unit tests for duration/. I can write more tests there if that's helpful 
Let me know if there's any other way I can help

icambron:
@CristiMacovei Yeah, so in general if you fix an issue, you write a test that would fail without your change and then make it pass with your code. That way, if someone make a change that breaks your fix, then they'll right away.

CristiMacovei:
Yeah, I'll get to that and I'll open another PR. Sorry, didn't know I was supposed to do that first

icambron:
It's ok for the test to be in the same PR, even if it only fails without your other code changes.

PR #1308, opened 2022-10-19
Fix Interval#count counting the endpoint as part of the interval

### LINKED ISSUE #1306: Interval#count incorrectly includes the end-point
**Describe the bug**
The Interval documentation states that
> An Interval object represents a **half-open** interval of time

To me this means that the interval includes the starting point, but not the end point. This is confirmed by checking `interval.contains(interval.end)`, which always returns false. However `Interval#count` also counts the endpoint.

**To Reproduce**
```js
const {Interval, DateTime} = require('luxon');

const startDate = DateTime.fromISO("2022-10-01T00:00:00.000+02:00");
const interval = Interval.after(startDate, {days: 2});

console.log(interval.contains(interval.end)); // as expected this is false
console.log(interval.count('days')); // this reports 3, but the interval only contains 2 calendar days (1st and 2nd)
console.log(interval.length('hours')); // as expected this reports exactly 48, because the first millisecond of the 3rd is not included
```

See here: https://runkit.com/embed/xxwq2ofg22di

`Interval#count` should not count the interval's endpoint.

- OS: Linux Mint
- Browser: Chrome 106, also happens on Node 14
- Luxon version: 3.0.4
- Your timezone: Europe/Berlin

**Additional context**
Looking at the code for `count`, it unconditionally adds 1 to the result of diffing start and end. This addition must only be done if `end` is actually different from `end.startOf(unit)`.

diesieben07:
Fixes #1306

One of the tests actually tested this incorrect behavior, which I have adjusted. I have also added a new test, explicitly testing the correct behavior.

### DIFF
--- a/src/interval.js
+++ b/src/interval.js
@@ -238,7 +238,7 @@ export default class Interval {
     if (!this.isValid) return NaN;
     const start = this.start.startOf(unit),
       end = this.end.startOf(unit);
-    return Math.floor(end.diff(start, unit).get(unit)) + 1;
+    return Math.floor(end.diff(start, unit).get(unit)) + (end.valueOf() !== this.end.valueOf());
   }
 
   /**
--- a/test/interval/info.test.js
+++ b/test/interval/info.test.js
@@ -54,9 +54,14 @@ test("Interval#count('years') returns 2 if the interval crosses the new year", (
   expect(i.count("years")).toBe(2);
 });
 
+test("Interval#count() does not count the endpoint of the interval", () => {
+  const i = DateTime.fromISO("2022-10-01").until(DateTime.fromISO("2022-10-03"));
+  expect(i.count("days")).toBe(2);
+})
+
 test("Interval#count() uses milliseconds by default", () => {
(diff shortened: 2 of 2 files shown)


icambron:
@diesieben07 Just a quick note to say I'm not ignoring this, just super slammed at the moment

nerdstep:
Just a head's up, this was a breaking change for us -- we had dealt with what we thought was a feature by subtracting 1 from the count. Not a huge deal since we caught it in unit tests, but I can imagine that this could potentially break other folks too.

diesieben07:
Sorry about that. There is always a fine line between "do we consider this a bug fix, or is this a breaking change". You don't want to go down the PHP route of maintaining stupid bugs like a decade later, because "it might break people's code", but you also don't want to intentionally break people's code.

nerdstep:
@diesieben07 it's all good, and I'm with you on that, I'd rather see a bug fixed than codified as an ugly wart. My only suggestion would perhaps be a stronger callout in the release notes. Either way, I certainly still appreciate the work and contributions towards maintaining this excellent library!

diesieben07:
I agree, we should work on improving the changelog for sure. Thank you for your feedback.

PR #1331, opened 2022-11-18
Fix tokenParse tests import broken by d6ea633

dobon:


dobon:
@icambron This fixes 2/3 of the tests I broke while adding Setting.twoDigitCutoffYear. I am not really sure why the ianaRegex error remains, though I can see that it is a result of including `import Settings from "../settings.js";` in `util.js`. Perhaps the relative paths are causing errors, since this would be ../../settings.js while in a test context? That's just a shot in the dark. I'm not really sure why the remaining test failure is occurring.

dobon:
@icambron I determined that the remaining test failure was a jest problem relating to how it handles named imports in certain situations. This was surfaced by a circular reference wherein jest's module loader incorrectly assigns the default export of ./util.js to each named export being imported in regexParser.js, rather than the named export itself. The circular reference (for lack of a better name):
```
english.test.js : import { formatRelativeTime } from "../../src/impl/english"
english.js : import { pick } from "./util.js"
util.js : import Settings from "../settings.js"
settings.js : import Locale from "./impl/locale.js"
locale.js : import DateTime from "../datetime.js"
datetime.js : import Duration from "./duration.js"
duration.js : import { parseISODuration, parseISOTimeOnly } from "./impl/regexParser.js"
regexParser.js : import { <various named exports> } from "./util.js";
```
 The second reference to './util' uses the cached default value from the previous reference, and fails to respect the named import.
 
 To work around the problem, and to avoid loading Settings everywhere that you might use Utils (and most of all to fix the red x on my pull request), I modified the util `untruncateYear` method so that it now has a second argument, `twoDigitCutoffYear`, and `Settings.twoDigitCutoffYear` is passed appropriately by the areas of code that use `untruncateYear`.

icambron:
I noticed that `ianaRegex` is only used by regexParser, so I fixed the circular reference by moving that. Seems cleaner. I cherry picked your first two commits and added my fix on top of that, which is here: #1336. I'm going to merge that and close this one.

PR #1340, opened 2022-12-03
find diff by duration hash instead of using repeated addition to cursor

dobon:


dobon:
I don't know why codecov/project is failing. It says -0.02% coverage, but the file goes from 100% coverage to 100% coverage...

icambron:
Awesome, this is how I was hoping this would turn out. This will go in 3.2.0 this week. Thanks!

PR #1343, opened 2022-12-12
docs(src-datetime): fix minor typo

baileypownell:
Just a quick fix for an extremely minor typo.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1926,7 +1926,7 @@ export default class DateTime {
 
   /**
    * Equality check
-   * Two DateTimes are equal iff they represent the same millisecond, have the same zone and location, and are both valid.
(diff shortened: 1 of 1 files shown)


icambron:
@baileypownell thanks. That isn't a typo, exactly, just perhaps unclear. "iff" is math speak for "if and only if" (see [here](https://en.wikipedia.org/wiki/If_and_only_if)). I'll admit that abbreviation isn't very friendly. Perhaps the most clarifying thing to do would be for us to simply spell out "if and only if"

baileypownell:
@icambron Ah, thanks for clarifying. I'll just close then! My bad.

icambron:
I went ahead and put in "if and only if" in f9917c4a. Thanks for bringing it to my attention

PR #1369, opened 2023-01-30
Fix support for Node.js 18.13+

### LINKED ISSUE #1364: DateTime.fromFormat fails in some cases when running Node 18.13.0
**Describe the bug**
I Node 18.13.0 parsing DateTime using the fromFormat method can in some cases fail parsing, where running in previous Node version it would not.

**To Reproduce**

```js
const { DateTime } = require("luxon");

const format = "M/d/yyyy tt";
const dateTimeString = "1/17/2023 11:00:00 PM";

const dateTime = DateTime.fromFormat(dateTimeString, format);

console.log(dateTime.isValid);
// "false" when running in 18.13.0
// "true" when running in 18.12.0

console.log(dateTime.toISO());
// "null" when running in 18.13.0
// "2023-01-17T23:00:00.000+01:00" when running in 18.12.0
```

I would expect the outcome of running the snippet to be the same in 18.12.0 and 18.13.0 and that the parsed DateTime is valid.
Instead in 18.12.0 the DateTime is correctly parsed while in 18.13.0 it is invalid.

- OS: Windows 11 (but originally noticed noticed in debian based docker container)
- Browser: Node 18.13.0
- Luxon version: 3.2.1
- Your timezone: Europe/Copenhagen

**Additional context**

mohd-akram:
Fixes #1364

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.9]
+        node-version: [18.13]
 
     steps:
     - uses: actions/checkout@v2
--- a/src/impl/formatter.js
+++ b/src/impl/formatter.js
@@ -47,6 +47,9 @@ export default class Formatter {
   }
 
   static parseFormat(fmt) {
+    // white-space is always considered a literal in user-provided formats
+    // the " " token has a special meaning (see unitForToken)
+
     let current = null,
       currentFull = "",
       bracketed = false;
--- a/src/impl/formatter.js
+++ b/src/impl/formatter.js
@@ -55,7 +55,7 @@ export default class Formatter {
       const c = fmt.charAt(i);
       if (c === "'") {
         if (currentFull.length > 0) {
-          splits.push({ literal: bracketed, val: currentFull });
+          splits.push({ literal: bracketed || /^\s+$/.test(currentFull), val: currentFull });
         }
         current = null;
         currentFull = "";
@@ -66,7 +66,7 @@ export default class Formatter {
         currentFull += c;
       } else {
         if (currentFull.length > 0) {
-          splits.push({ literal: false, val: currentFull });
+          splits.push({ literal: /^\s+$/.test(currentFull), val: currentFull });
         }
         currentFull = c;
         current = c;
@@ -74,7 +74,7 @@ export default class Formatter {
     }
 
     if (currentFull.length > 0) {
-      splits.push({ literal: bracketed, val: currentFull });
+      splits.push({ literal: bracketed || /^\s+$/.test(currentFull), val: currentFull });
     }
 
     return splits;
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -180,6 +180,8 @@ function unitForToken(token, loc) {
         // because we don't have any way to figure out what they are
         case "z":
           return simple(/[a-z_+-/]{1,256}?/i);
+        case " ":
+          return simple(/\s/);
         default:
           return literal(t);
       }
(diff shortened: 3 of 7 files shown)


cscrosati:
/cc @icambron 
It appears this fix could help folks that depends on Luxon for date format parsing in Chrome/Edge v110+ and Node v18.13+.

Can it be merged and released as a minor update if you agree?

diesieben07:
I don't think this should be done always, it should be opt-in. An option like `looseWhitespace` for `fromFormat` would be good.
Additionally, I don't think we want to allow new-lines (which `\s` covers).

diesieben07:
Some clarification of the above:

- By default `fromFormat` should keep behaving as it does now.
- An option should be added to `fromFormat` like `looseWhitespace`, which toggles an alternative behavior. In this alternative behavior, any whitespace given by the unterlying system for the "macro tokens" (e.g. `tt` for "local time")  is mapped to `[^\S\n\r]` (match any whitespace except new lines). This way all whitespace is "treated equal" in terms of parsing.
- I don't think _outputting_ of dates should be affected. The output of `toLocaleString` is _by definition_ variable. If you need a consistent output, use `toFormat`.

diesieben07:
> The docs actually say t is for localized time even when parsing, so it should be able to parse the localized output without any extra options.

Correct, and so it does. But the localized output in this case contains a narrow width space, not a normal space. As such a normal space _does not fit the format_. This change makes parsing more lax, allowing any whitespace in place of where the localized time says a narrow width space should be.

> Also, new Date accepts a newline as space when passing it a localized string, so it makes sense to match that behavior.

`new Date()` or more specifically `Date.parse` only guarantees support for ISO-8601. It will _try_ to parse other formats, but doing so is implementation dependent and may not work at all. Attempting to parse a localized date with this method is futile (is `03-04-2022` March 4th or April 3rd?) and not what this method is for.

diesieben07:
To clarify:
```js
const dt = DateTime.now();
const roundtrip = DateTime.fromFormat(dt.toLocaleString(DateTime.TIME_SIMPLE), 't');
```
This always works. `t` parses whatever `toLocaleString(DateTime.TIME_SIMPLE)` would do - and that is _implementation dependent_.

mohd-akram:
> This change makes parsing more lax, allowing any whitespace in place of where the localized time says a narrow width space should be.

It still is a regular space in certain runtimes. Localized time is by definition lax. If one needs strict parsing they should use explicit tokens. Otherwise, the `t` token is virtually useless without your specified option since most will expect it to parse a regular space - and [browsers were patched to do so](https://bugs.chromium.org/p/v8/issues/detail?id=13494).

diesieben07:
I'm sorry. I think I was running down a path of "but the docs say so, so it must be done this way".
Thinking about it more broadly, you are correct. When parsing, white-spaces should be treated equally and in fact Luxon already does something similar.

One change I would suggest though is to make the regex not accept new lines, i.e. use `[^\S\n\r]` instead of `\s`. I don't think it makes sense to allow parsing
```
10:30
PM
```

diesieben07:
Another note, I am not a fan of how special-cased this implementation is. I think it would be better, to handle the white-space normalization similar to how it is already done for the other parsing code in `tokenParser.js`. The `literal` method should be adjusted to normalize white-space, like `oneOf` already does.

diesieben07:
Thank you for your contribution @mohd-akram and sorry for the back and forth.

dasa:
It [seems](https://github.com/nodejs/node/issues/46123#issuecomment-1428529613) Node and the browsers are reverting the breaking change, but now at least Luxon is compatible either way 😄

ranjan-purbey:
Would it be possible to backport this fix for luxon 1.x as well?

icambron:
@ranjan-purbey no, pretty much not going to make Luxon 1.0 fixes. Note that upgrading is not usually very involved, so I encourage you to do that instead.

Issue #1363, opened 2023-01-16
Custom Timezones only work in certain circumstances


Munksey:
**Describe the bug**

I'm not entirely sure if this is a bug or if this isn't intended to be a supported feature, but it seems like custom timezones are poorly supported (or at the very least, poorly documented). Particularly, there are a few things I'd like to point out about the behavior:

1. There are discrepancies between when the `isUniversal` flag is `true` or `false`.
2. There are discrepancies between when a custom timezone is overriding an existing system timezone.

I did get the impression that custom timezones should be supported from reading [this blurb](https://github.com/moment/luxon/blob/857812f9ccb1351b4f7cd7f90f3f9e5e17d7fe94/docs/zones.md#specifying-a-zone) about specifying a zone.

> A custom implementation of Luxon's Zone interface (advanced only)

At the very least, I do think that if something comes out of this issue, it should be at least to update the documentation and perhaps add an example of how to implement custom timezones and their limitations. Or, to completely remove the blurb all together.

**To Reproduce**

Below is a basic set up for a custom timezone. In this case the behavior for `isUniversal` is not actually honored, all times will be offset by one hour after `3600000`.

There are four supported timezones, two of which are overrides: `Bogus/NotUniversal`, `Bogus/Universal`, `America/New_York`, `America/Los_Angeles`.

```
const { DateTime, FixedOffsetZone, Zone } = require('luxon');

class MyZone extends Zone {
  /**
   * Determines whether or not the time zone is universal.
   */
  static UNIVERSAL = {
    'Bogus/NotUniversal': false,
    // Overrides a known time zone.
    'America/New_York': false,
    'Bogus/Universal': true,
    // Overrides a known time zone.
    'America/Los_Angeles': true,
  }

  constructor(name) {
    super();
    this._name = name;
  }

  get isUniversal() {
    return MyZone.UNIVERSAL[this._name];
  }

  get isValid() {
    return true;
  }

  get name() {
    return this._name;
  }

  get type() {
    return 'custom';
  }

  equals(zone) {
    return zone.name === this.name;
  }

  offset(ms) {
    const ONE_HOUR = 3600000;
    // After January 1, 1970 01:00 UTC, set the clocks forward an hour.
    return ms >= ONE_HOUR ? 60 : 0;
  }

  offsetName() {
    return 'N/A';
  }

  formatOffset(...args) {
    return FixedOffsetZone.prototype.formatOffset(...args);
  }
}
```

**Actual vs Expected behavior**

[...]

PR #1377, opened 2023-02-09
Custom zone formatting support

thatsmydoing:
This addresses https://github.com/moment/luxon/issues/1363

This addresses a couple of bugs related to non-IANA timezone formatting
 * timezones can be displayed even if `timeZoneName` is not set and partial hour fixed-offset zones don't handle this properly
 * non-universal custom non-IANA zones are treated as IANA zones and are improperly formatted
 * allow all zones to have timezone in the format using `formatToParts`

### DIFF
--- a/src/impl/locale.js
+++ b/src/impl/locale.js
@@ -1,4 +1,4 @@
-import { padStart, roundTo, hasRelative } from "./util.js";
+import { padStart, roundTo, hasRelative, formatOffset } from "./util.js";
 import * as English from "./english.js";
 import Settings from "../settings.js";
 import DateTime from "../datetime.js";
@@ -203,7 +203,7 @@ class PolyDateFormatter {
     if (this.opts.timeZone) {
       // Don't apply any workarounds if a timeZone is explicitly provided in opts
       this.dt = dt;
-    } else if (dt.zone.isUniversal) {
+    } else if (dt.zone.type === "fixed") {
       // UTC-8 or Etc/UTC-8 are not part of tzdata, only Etc/GMT+8 and the like.
       // That is why fixed-offset TZ is set to that unless it is:
       // 1. Representing offset 0 when UTC is used to maintain previous behavior and does not become GMT.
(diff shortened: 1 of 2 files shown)


src/impl/locale.js
```diff
@@ -212,25 +216,23 @@ class PolyDateFormatter {
...
-        // So we have to make do. Two cases:
-        // 1. The format options tell us to show the zone. We can't do that, so the best
-        // we can do is format the date in UTC.
-        // 2. The format options don't tell us to show the zone. Then we can adjust them
-        // the time and tell the formatter to show it to us in UTC, so that the time is right
-        // and the bad zone doesn't show up.
+        // Not all fixed-offset zones like Etc/+4:30 are present in tzdata so
+        // we manually apply the offset and substitute the zone as needed.
         z = "UTC";
-        if (opts.timeZoneName) {
```
thatsmydoing, line 223: This isn't entirely correct since using `timeStyle: "full"` will also show the timezone.

src/impl/locale.js
```diff
@@ -212,25 +216,23 @@ class PolyDateFormatter {
...
-          this.dt = dt;
-        } else {
-          this.dt = dt.offset === 0 ? dt : DateTime.fromMillis(dt.ts + dt.offset * 60 * 1000);
```
thatsmydoing, line 226: This throws away any locale information so it would be more correct to just switch the datetime to UTC and apply the offset there.

src/impl/locale.js
```diff
@@ -239,11 +241,35 @@ class PolyDateFormatter {
...
 
   formatToParts() {
-    return this.dtf.formatToParts(this.dt.toJSDate());
+    const parts = this.dtf.formatToParts(this.dt.toJSDate());
+    if (this.originalZone) {
+      return parts.map((part) => {
+        if (part.type === "timeZoneName") {
+          const offsetName = this.originalZone.offsetName(this.dt.ts, {
+            locale: this.dt.locale,
+            format: this.opts.timeZoneName,
```
thatsmydoing, line 261: The docs currently say the format allows `short` or `long` but `Intl.DateTimeFormat` now supports more options. I think it's more correct to just pass it through as the `IANAZone` technically allows you to pass any of the valid options. To be more correct, we might also want to consider adding `timeStyle` as an option since that can also influence the timezone output.

thatsmydoing:
Theoretically, the logic for fixed offset timezones can simply reuse the one for custom zones but I thought that might be too much of a breaking change compared to the current approach of using IANA zones for whole hour offsets.

icambron:
@thatsmydoing this is great. I think we can just merge this as-is. Thanks for the contribution!

PR #1385, opened 2023-02-20
Update package.json

DiviBretschneider:
Add version prop to src/package.json as it is mandatory for Webpack 5/Module Federation.


icambron:
I applied this manually and released it in 3.3.0

PR #1391, opened 2023-03-06
Correctly handle hourCycle when expanding macro tokens

### LINKED ISSUE #1197: Issue at MUI: AdapterLuxon doesn't provide placeholder text (hh:mm / dd.mm.yy) in Date- or TimePickers
Hello, first I opened a bug at Material-UI for this issue. As you can see, it seems to be in your responsibility:
[MaterialUI-Issues](https://github.com/mui/mui-x/issues/4755)

**Describe the bug**
I wrapped TimePicker and also DatePicker in my app with LocalizationProvider.
Both gets the prop: dateAdapter={AdapterLuxon} and locale={string} ("en-US" / "de-DE" ...).

The am/pm behaviour works as expected, but I never get any placeholder text.
I switched from moment library (with AdapterMoment in LocalizationProvider) to Luxon DateTime now and the placeholder disappears.
Will say, with Moment it worked well and also as we can see with DateFNS it work's also like on your documentation:
[LocalizationProvider](https://mui.com/components/time-picker/#localization)

**To Reproduce**
[CodeSandbox](https://codesandbox.io/s/localizationprovider-issue-6yepj?file=/src/App.tsx)

Placeholder textes should also appear with AdapterLuxon in all Date- / Timepickers.

- OS: Win10 / MacOs 12.3.1
- Browser Chromium Browsers: Edge: Version 100.0.1185.50 (Offizielles Build) (arm64) / Brave: [Version 1.38.109 Chromium: 101.0.4951.41 (Offizieller Build) (arm64)] / ...
- Luxon version 2.3.0 in cloud app / cadesandbox: 2.3.2
- Your timezone: Berlin

**Additional context**
This problem is a UX requirement in order to quickly and easily convey to the user what needs to be entered into the field. Additionally with the info in which format.

diesieben07:
When expanding macro tokens, we must expand "hour" either into "h" or "H", depending on the desired hour cycle. 
This fixes #1197.

I have not changed the parsing of `h`, which currently accepts 24 hours just the same as `H`. This would be a breaking change and I am not sure it is worth introducing another special case in the parser for this.

### DIFF
--- a/src/impl/formatter.js
+++ b/src/impl/formatter.js
@@ -101,24 +101,25 @@ export default class Formatter {
     return df.format();
   }
 
-  formatDateTime(dt, opts = {}) {
-    const df = this.loc.dtFormatter(dt, { ...this.opts, ...opts });
-    return df.format();
+  dtFormatter(dt, opts = {}) {
+    return this.loc.dtFormatter(dt, { ...this.opts, ...opts });
   }
 
-  formatDateTimeParts(dt, opts = {}) {
-    const df = this.loc.dtFormatter(dt, { ...this.opts, ...opts });
-    return df.formatToParts();
+  formatDateTime(dt, opts) {
+    return this.dtFormatter(dt, opts).format();
   }
 
-  formatInterval(interval, opts = {}) {
-    const df = this.loc.dtFormatter(interval.start, { ...this.opts, ...opts });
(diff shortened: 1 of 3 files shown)


flaviendelangle:
Hi,

Thanks again for your work !
Do you have a time estimate for the release ?

PR #1398, opened 2023-03-17
Fix DateTime#diff for cross-zone diffs by checking where dayDiff is off by 2 days (#1165)

### LINKED ISSUE #1165: Duration.diff returns negative number

ConnorLanglois:
Fixes #1165 (I think). Please review code and see if correct; I am not a datetime expert and this date stuff can be hard.

### DIFF
--- a/src/impl/diff.js
+++ b/src/impl/diff.js
@@ -34,6 +34,13 @@ function highOrderDiffs(cursor, later, units) {
 
       if (highWater > later) {
         results[unit]--;
+
+        const highWater2 = earlier.plus(results);
+        if (highWater2 > later) {
+          results[unit]--;
+          highWater = highWater2;
+        }
+
         cursor = earlier.plus(results);
       } else {
         cursor = highWater;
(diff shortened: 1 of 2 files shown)


diesieben07:
Thank you for your contribution, I have merged a solution based on your code and your test manually.

PR #1400, opened 2023-03-21
SQL time support trim milliseconds

dreak:
A more convenient way to get no milliseconds of sql time.


icambron:
Oops, i merged this prematurely. I need to make sure the options are consistent with other methods.

PR #1404, opened 2023-03-24
Fix Duration#toISOTime depending on the locale

### LINKED ISSUE #1403: Duration.toISOTime has locale-dependent output (using ar-QA locale)
**Describe the bug**
We ran into a production error today trying to parse an ISO time passed from the client to the server. The client had an Arabic locale and for some reason the locale-formatted time value was making its way to the server and crashing the ISO parse.

**To Reproduce**
I eventually boiled down the problem to this repro (running in the console at https://moment.github.io/luxon) :
```javascript
> luxon.Settings.defaultLocale = "ar-QA"
'ar-QA'
> luxon.Duration.fromObject({hours: 1, minutes: 4}).toISOTime()
'٠١:٠٤:٠٠.٠٠٠'
> luxon.DateTime.now().toISOTime()
'16:53:34.372-04:00'
> luxon.Duration.fromISOTime(luxon.Duration.fromObject({hours: 1, minutes: 1}).toISOTime()).invalidExplanation
`the input "٠١:٠١:٠٠.٠٠٠" can't be parsed as ISO 8601`
```

`luxon.Duration.fromObject({hours: 1, minutes: 4}).toISOTime()` outputs `٠١:٠٤:٠٠.٠٠٠` in certain locales.
I would have expected the toISOTime to be locale-independent and always output `01:04:00.000`

If I'm misunderstanding the ISO standard in this case, I would at least expect the roundtrip not to fail:
```javascript
> luxon.Duration.fromISOTime(luxon.Duration.fromObject({hours: 1, minutes: 1}).toISOTime()).invalidExplanation
`the input "٠١:٠١:٠٠.٠٠٠" can't be parsed as ISO 8601`
```

**Desktop:**

- OS: Windows, Mac
- Browser: Chrome 111
- Luxon version: 3.3.0
- Your timezone: "America/New_York"

diesieben07:
I have created this as a pull request, because _technically_ this is a breaking change.
Previously `DateTime#toISOTime` would ignore milliseconds even if only `suppressSeconds` was set. `Duration#toISOTime` did not do so. I have aligned the behavior between the two, so that no data is ever silently ignored (suppressing only works if _all_ hidden units are 0).

Please let me know if you think preserving the original behavior of suppressSeconds on DateTime is worthwhile. 

Fixes #1403

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -224,14 +224,14 @@ function toISOTime(
   if (extended) {
     c += ":";
     c += padStart(o.c.minute);
-    if (o.c.second !== 0 || !suppressSeconds) {
+    if (o.c.millisecond !== 0 || o.c.second !== 0 || !suppressSeconds) {
       c += ":";
     }
   } else {
     c += padStart(o.c.minute);
   }
 
-  if (o.c.second !== 0 || !suppressSeconds) {
+  if (o.c.millisecond !== 0 || o.c.second !== 0 || !suppressSeconds) {
     c += padStart(o.c.second);
 
     if (o.c.millisecond !== 0 || !suppressMilliseconds) {
--- a/src/duration.js
+++ b/src/duration.js
@@ -12,6 +12,7 @@ import {
   roundTo,
 } from "./impl/util.js";
 import Settings from "./settings.js";
+import DateTime from "./datetime.js";
 
 const INVALID = "Invalid Duration";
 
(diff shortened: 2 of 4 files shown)


icambron:
That is technically breaking, but only in the sense that all changes are technically breaking, because of Hyrum's Law. I say we merge this. (Also the code looks good)

PR #1412, opened 2023-04-02
Truncate milliseconds shorthand in ISO

EzeGmnz:
Adds a new parameter to DateTime#toISO and toISOTime called truncateMilliseconds which will remove the milliseconds from the format, irrespective of their value.

Added tests

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -217,24 +217,33 @@ function toISOTime(
   extended,
   suppressSeconds,
   suppressMilliseconds,
+  truncateMilliseconds,
   includeOffset,
   extendedZone
 ) {
   let c = padStart(o.c.hour);
+
+  const includeMilliseconds =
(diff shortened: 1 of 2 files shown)


erdnaxeli:
Hi, any news about this? this would be very helpful. Thanks!

BalusC:
I see this is still not merged while this is certainly useful.

For the time being, this can be work arounded as follows:
```
dateTime.startOf("second").toISO({ suppressMilliseconds: true });
```

dobon:
> What is the status of this issue?

This functionality is added by https://github.com/moment/luxon/pull/1670 which will be released on the next release of luxon (soon, I think?). It adds an optional `precision` parameter for `toISO`, which you can configure to truncate milliseconds reliably.

This PR can be closed as duplicate.

NogaMan:
When is the next release of luxon?

PR #1415, opened 2023-04-03
docs: Note the new TS option for throwOnInvalid.

CarsonF:
I am adding support for the invalid return types in the TypeScript definitions.
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64995
With this I added a way to opt-in to the stricter types when `throwOnInvalid` is enabled.
I'm documenting this here for users, since we don't really have another documentation place.

### DIFF
--- a/docs/validity.md
+++ b/docs/validity.md
@@ -75,6 +75,16 @@ DateTime.now().setZone("America/Blorp"); //=> Error: Invalid DateTime: unsupport
 
 You can of course leave this on in production too, but be sure to try/catch it appropriately.
 
+For TypeScript users, this runtime setting narrows the return types. You need manually inform TypeScript of this setting with this snippet.
+```ts
+declare module 'luxon' {
+  interface TSSettings {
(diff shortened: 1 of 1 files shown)


pvogel1967 (commented):
This doc is inadequate.  Should be clear this should be in a local types/luxon/index.d.ts file and it should `import { TSSettings } from 'luxon'` to ensure the module declare here is merged with the main types module from DefinitelyTyped.

CarsonF:
> This doc is inadequate. Should be clear this should be in a local types/luxon/index.d.ts file and it should `import { TSSettings } from 'luxon'` to ensure the module declare here is merged with the main types module from DefinitelyTyped.

No this does not have to be in locally configured "types" folder. It should be done in the same spot as you enable the runtime setting, as I thought the doc change communicated by being right under that runtime setting. That setting has to be executed early on in your application, and doing so means TS picks it up and includes it (merges the interface correctly).

PR #1427, opened 2023-04-30
Fix: Remove "TTT" from valid parsing tokens

6ixfalls:
One-liner change. Spent a while trying to figure out why I couldn't use fromFormat with `TTT`, and ended up being not supported by the function even though it's in the parsing table of valid tokens, which may be confusing, albeit also already stating that `Luxon also doesn't support the "macro" tokens that include offset names, such as "ttt" and "FFFF"`.


icambron:
Ah. sorry about that. Thanks for the fix

PR #1435, opened 2023-05-13
refactor: remove unused `defaultOK` parameter

MonstraG:
`listingMode` function was being passed `defaultOK` to, but this function doesn't accept any arguments since 22/12/2021 (c96b8132).


icambron:
@MonstraG thanks!

PR #1437, opened 2023-05-20
Fix duck test to match Zone interface

### LINKED ISSUE #1426: Duck check fails in normalizeZone
**Describe the bug**
In my scenario, I believe I have a date whose zone instance came from a separate node module, and thus the `instanceof Zone` check fails.  I see you have duck checking in the `normalizeZone` function to handle this use-case.  

However, it appears to be written with the wrong assumption that `offset` is a getter property, whereas it seems actually to be [defined as a **function** in Zone.js](https://github.com/moment/luxon/blob/master/src/zone.js#L63-L71).  This causes the check to fail unexpectedly.

image

justinbhopper:
Fixes #1426

`normalizeZone` uses `offset` on a Zone for the duck test, but incorrectly assumes it is a property that returns a number when in fact it is a function/method on Zone.

The duck test is particularly important for scenarios where package dependencies brought in their own copy of luxon (perhaps due to version mismatches), causing the `instance of` check to fail.  The duck test helps ensure compatibility with other instances of luxon modules.

### DIFF
--- a/src/impl/zoneUtil.js
+++ b/src/impl/zoneUtil.js
@@ -24,7 +24,7 @@ export function normalizeZone(input, defaultZone) {
     else return FixedOffsetZone.parseSpecifier(lowered) || IANAZone.create(input);
   } else if (isNumber(input)) {
     return FixedOffsetZone.instance(input);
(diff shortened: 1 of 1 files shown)


icambron:
@justinbhopper thanks

PR #1454, opened 2023-06-22
Implement localized week information

### LINKED ISSUE #1447: Support Intl start of week settings
I've seen several closed issues regarding this problem where there is no way to set the start of week manually or based on the new Intl `getWeekInfo` method. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo 
I am working on an international application which will need dynamic switching of startOfWeek for the appropriate audience. We can capture this via the new Intl API, but I could not find anywhere to either control it in luxon or integrating the new functionality into setting locale. 

Setting locale should leverage the Intl `getWeekInfo` API to set the startOfWeek weekday to the appropriate day.

Add a config or setting option to change the startOfWeek to a weekday (0-6)

**Additional context**
Some component libraries do not let devs manually change the start of week, it is based on the datetime library.
https://github.com/mui/material-ui/issues/30591

Intl API:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo

diesieben07:
Use [`Intl.Locale#getWeekInfo`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo) to provide locale-dependent week information, such as the start of the week.

Fixes #1447

### Current status:

- [x] `Info.features` reports on whether localized week info is available in the current environment
- [x] `Info.getStartOfWeek`: returns 1 (Monday) through 7 (Sunday) for first day of the week according to locale
- [x] `Info.getMinimumDaysInFirstWeek` for getting minimum days in the first week of the year
- [X] `Info.getWeekendWeekdays`  for getting weekend days
- [x] `DateTime#isWeekend`: returns whether the current date is on the weekend according to the locale
- [x] `DateTime#localWeekday`: returns 1-7 for first to last day according to locale
- [x] `DateTime#localWeekNumber` and `DateTime#localWeekYear`; same as weekNumber and weekYear, but respecting the locale. I need to understand the math here before I can implement it.
- [x] `DateTime#hasSame`, `DateTime#startOf`, `DateTime#endOf` and `Interval#count` accept a new option `useLocaleWeeks`, when enabled the `week` unit will act according to the locale week information
- [x] Tests
- [x] (Potentially) Caching for the localized week data (weekNumber, weekYear)
- [X] Investigate how to let the user override the defaults. There are BCP tags to set the start of the week, but browsers seem to ignore them, so we probably need to provide our own configuration option, similar to `numberingSystem`.
- [X] Accept locale-based week units in `DateTime.fromObject` and `DateTime#set`
- [X] Add `DateTime#weeksInLocalWeekYear`, analogous to `weeksInWeekYear`
- [x] Investigate supporting formatting tokens for `localWeekNumber` and `localWeekYear`.
- [X] Documentation

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1493,6 +1493,8 @@ export default class DateTime {
   /**
    * "Set" this DateTime to the beginning of a unit of time.
    * @param {string} unit - The unit to go to the beginning of. Can be 'year', 'quarter', 'month', 'week', 'day', 'hour', 'minute', 'second', or 'millisecond'.
+   * @param {Object} opts - options
+   * @param {boolean} [opts.useLocaleWeeks=false] - If true, use weeks based on the locale, i.e. use the locale-dependent start of the week
    * @example DateTime.local(2014, 3, 3).startOf('month').toISODate(); //=> '2014-03-01'
    * @example DateTime.local(2014, 3, 3).startOf('year').toISODate(); //=> '2014-01-01'
    * @example DateTime.local(2014, 3, 3).startOf('week').toISODate(); //=> '2014-03-03', weeks always start on Mondays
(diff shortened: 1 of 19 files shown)


src/datetime.js
```diff
@@ -1129,6 +1130,30 @@ export default class DateTime {
...
+   * Get the of the week according to the locale.
+   * 0 is the first day of the week and 6 is the last day of the week.
+   * If the locale assigns Monday as the first day of the week, then a date which is a Monday will return 0,
+   * @returns {number}
+   */
+  get localDayOfWeek() {
+    return this.isValid ? isoWeekdayToLocal(this.weekday, this.loc.getStartOfWeek()) : NaN;
+  }
+
+  get localeWeekNumber() {
```
icambron, line 1151: You used `locale` in the name here and `local` for the name in `localDayOfWeek`. We should be consistent. I prefer `local`

icambron:
Looks really good. Left one comment

diesieben07:
After a lot of work, this is now ready for review. All features I wanted to add have been implemented, including tests.

icambron:
This is great. Merging as-is.

PR #1464, opened 2023-07-09
fix: Info.months for Islamic calendar

adotnusiyan:
`Info.months()` for Islamic has duplicate month "**Rabiʻ I**" and also Islamic months' order is incorrect.

Fixed by changing the year from 2016 to 2009 which is the most recent year that Islamic and Gregorian months are sync.

[Islamic months and their order](https://en.wikipedia.org/wiki/Islamic_calendar#Months)


icambron:
@adotnusiyan thanks

PR #1466, opened 2023-07-13
Default config DateTime constructor

jmostaer:
When using DateTime with mappers such as class-transfomers, these libraries will try to use the DateTime constructor without parameters. This currently will throw an exception. Setting a default value for config in the constructor solves this problem.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -432,7 +432,7 @@ export default class DateTime {
   /**
    * @access private
    */
-  constructor(config) {
+  constructor(config = {}) {
     const zone = config.zone || Settings.defaultZone;
 
     let invalid =


diesieben07:
The constructor is private API and should not be called by anyone but Luxon itself.
Can you clarify which library you are using that requires calling the constructor directly?

jmostaer:
We are using class-transformer. I think class-transformer uses the constructor to try to create a deep copy of the DateTime instance.

diesieben07:
I haven't used class-transformer myself.
Looking at their docs, it seems they show how to add manual [transformations](https://github.com/typestack/class-transformer#additional-data-transformation). This (or similar APIs) should be used to make the library not use the private constructor.

jmostaer:
We do use transformations. In the class we are tranforming to, the date is not even stored as a luxon DateTime, but as string. Luxon is only used in the plain object we are trying to convert to the class. However, to my understanding, class-transformer will first try to create a deep copy of the object to be transformed before executing the transformations on the object (probably to avoid accidentally altering the original object in the transformations). This is where the constructor of DateTime is used an where the problem above pops up.

PR #1470, opened 2023-07-18
Update datetime.js documentation

pixeloution:
the @param as {DateTime} was incorrect and while it technically accepts {string|Invalid}, looking at the code and original documentation lended the impression you'd prefer a string argument for the public API docs


icambron:
@pixeloution thanks

Issue #1482, opened 2023-08-13
Duration subtraction regression


Maxim-Mazurok:
**Describe the bug**
Previously subtracting "P1W" from "P14D" resulted in "P1W" which makes sense. Now it's "P0.651W2DT10H17M45.6S"

**To Reproduce**
You can run this at https://npm.runkit.com/luxon
```js
const { DateTime, Duration } = require("luxon@3.4.0");

const purchaseDate = DateTime.fromObject(
  { day: 1, month: 1, year: 2020 },
  { zone: "UTC" }
);
const warranty = Duration.fromObject(
  { days: 14 },
  { conversionAccuracy: "longterm" }
);
const now = DateTime.fromObject(
  { day: 8, month: 1, year: 2020 },
  { zone: "UTC" }
);
const age = now.diff(purchaseDate).shiftToAll();
const warrantyLeft = warranty.minus(age).shiftToAll();

console.log({
  purchaseDate: purchaseDate.toString(),
  warranty: warranty.toString(),
  now: now.toString(),
  age: age.toString(),
  warrantyLeft: warrantyLeft.toString(),
});
```
it will produce 
```
purchaseDate: "2020-01-01T00:00:00.000Z"
warranty: "P14D"
now: "2020-01-08T00:00:00.000Z"
age: "P1W"
warrantyLeft: "P0.651W2DT10H17M45.6S"
```

**Actual vs Expected behavior**
Running luxon 3.3.0 at https://npm.runkit.com/luxon:
```js
const { DateTime, Duration } = require("luxon@3.3.0");

const purchaseDate = DateTime.fromObject(
  { day: 1, month: 1, year: 2020 },
  { zone: "UTC" }
);
const warranty = Duration.fromObject(
  { days: 14 },
  { conversionAccuracy: "longterm" }
);
const now = DateTime.fromObject(
  { day: 8, month: 1, year: 2020 },
  { zone: "UTC" }
);
const age = now.diff(purchaseDate).shiftToAll();
const warrantyLeft = warranty.minus(age).shiftToAll();

console.log({
  purchaseDate: purchaseDate.toString(),
  warranty: warranty.toString(),
  now: now.toString(),
  age: age.toString(),
  warrantyLeft: warrantyLeft.toString(),
});
```
results in 
```
purchaseDate: "2020-01-01T00:00:00.000Z"
warranty: "P14D"
now: "2020-01-08T00:00:00.000Z"
age: "P1W"
warrantyLeft: "P1W"
```
[...]

PR #1489, opened 2023-08-16
Handle invalid Durations in toHuman and toMillis

### LINKED ISSUE #1488: Breaking regression from 3.3.0 to 3.4.0 in milliseonds
**Describe the bug**
I use Luxon for nodevu, which is a module to get data about Node.js versions. I run nightly builds of static data for the module, and they started failing a bit ago. I recently had a user reach out saying they'd realized there was an issue with Luxon. Upon pretty basic investigation, it seems there's an issue between Luxon 3.3.0 and 3.4.0 where behavior around milliseconds seemingly changed.

**To Reproduce**
Please share a minimal code example that triggers the problem:
I wasn't able to narrow down _what_ in Luxon is triggering this, but I did create a minimal reproduction to trigger the issue with nodevu. I apologize for not being able to dig in more and make something more minimal, but I'm currently OOO and am limited on time to work on code.

```bash
git clone https://github.com/bnb/luxon-breaking-minor.git
npm install
npm start
```

That will output nodevu output that can safely be ignored _and_ the error + stack trace.

Milliseconds management should not change in minors.

- OS: [e.g. iOS] macOS
- Browser [e.g. Chrome 84, safari 14.0] N/A
- Luxon version [e.g. 1.25.0] 3.4.0
- Your timezone [e.g. "America/New_York"] America/New_York

diesieben07:
Fixes #1488 
I've also added tests so that this behavior is no longer broken in the future.

@icambron It would be good to get this out as 3.4.1 relatively quickly, considering this is a regression.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -462,6 +462,8 @@ export default class Duration {
    * ```
    */
   toHuman(opts = {}) {
+    if (!this.isValid) return INVALID;
+
     const l = orderedUnits
       .map((unit) => {
         const val = this.values[unit];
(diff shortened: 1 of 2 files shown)


icambron:
@diesieben07 thanks. I'd also like to fix #1482. I'll see if I can get to that now.

icambron:
Released in 3.4.1

PR #1494, opened 2023-08-26
Fix and improve `Duration#normalize`

### LINKED ISSUE #1493: Regression: Duration.normalize() not working properly after 3.4.1
**Describe the bug**
See https://github.com/moment/luxon/actions/runs/5943754818/job/16119553727

    - Expected  - 2
    + Received  + 2

      Object {
    -   "days": 363,
    -   "years": 1,
    +   "days": -2,
    +   "years": 2,
      }

    - Expected  - 2
    + Received  + 2

      Object {
    -   "days": 2,
    -   "months": 0,
    +   "days": -28,
    +   "months": 1,
      }
Likely the fix for #1482 in 3.4.1 (https://github.com/moment/luxon/commit/11e454c0fdd2da8cc6b5821b89c03c52a7c64b91) broke #1467 

**To Reproduce**
`Duration.fromObject({ hours: 96, minutes: 0, seconds: -10 }).normalize().toObject())`
should equal `{ hours: 95, minutes: 59, seconds: 50}` but no conversion happened.

All tests should be passing

diesieben07:
Fixes #1493

In the process of fixing this, I have updated the algorithm used in `normalizeValues` to make it more clear what it actually does and improved the documentation for `Duration#normalize` to clarify what "canonical representation" means.

Additionally, I have removed the broken implementation of `convert` that was introduced in #1467. After the refactoring of `normalizeValues`, `Duration#shiftTo` was the only user of `convert`, and unnecessarily so. shiftTo used convert to perform some of the same steps that normalize already does, which shiftTo calls afterwards. As such `convert` has been removed.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -143,11 +143,35 @@ function convert(matrix, fromMap, fromUnit, toMap, toUnit) {
 }
 
 // NB: mutates parameters
-function normalizeValues(matrix, vals) {
+function normalizeValues(matrix, vals, negative) {
+  const factor = negative ? -1 : 1;
+  // the logic below assumes the overall value of the duration is positive
+  // if this is not the case, factor is used to make it so
+
   reverseUnits.reduce((previous, current) => {
     if (!isUndefined(vals[current])) {
       if (previous) {
-        convert(matrix, vals, previous, vals, current);
+        const previousVal = vals[previous] * factor;
+        if (previousVal < 0) {
+          // lower order unit is negative (e.g. { years: 2, days: -2 })
+          // normalize this by reducing the higher order unit by the appropriate amount
+          // and increasing the lower order unit
+          // this can never make the higher order unit negative, because this function only operates
+          // on positive durations, so the amount of time represented by the lower order unit cannot
+          // be larger than the higher order unit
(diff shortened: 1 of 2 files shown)


icambron:
@diesieben07 i went ahead and merged this, but i'm a little concerned that there are no new tests

diesieben07:
@icambron I looked at the existing tests and did not find any cases that were not covered.
- #1493 was already covered by tests (3.4.1 was released with broken tests).
- # 1467 added tests already, which still pass.

However I can take a look again and try to cover more cases if you think it is important.

icambron:
Ah, I'd missed that we had a broken test. That's because my local env has a few failing tests because I use a newer version of node (causes the format strings to not match up), and I think the real test failure got lost in it. I should have noticed in the CI though. I updated the test to use a newer Node.

s-palmer:
Looks like this PR may break builds that use webpack 4 similarly to this: https://github.com/alpinejs/alpine/pull/3679

Changing this line: `let sum = vals.milliseconds ?? 0;`
to: `let sum = vals.milliseconds !== null && vals.millisecondsl !== undefined ? vals.milliseconds : 0`

resolves the build issues

PR #1498, opened 2023-09-03
Fix Duration#normalize, shiftTo and shiftToAll

### LINKED ISSUE #1496: shiftAll() regression
**Describe the bug**
What used to be `P99Y11M3W2DT10H29M6S` in v3.3.0 is now `P99Y11M3.3481250000000005W` in v3.4.2

**To Reproduce**
```ts
const { DateTime, Duration } = require("luxon@3.4.2");

const purchaseDate = DateTime.fromObject(
  { day: 1, month: 1, year: 2020 },
  { zone: "UTC", conversionAccuracy: "longterm" }
);
const warranty = Duration.fromObject({ years: 100 }, { zone: "UTC", conversionAccuracy: "longterm" });
const now = DateTime.fromObject(
  { day: 8, month: 1, year: 2020 },
  { zone: "UTC", conversionAccuracy: "longterm" }
);
const age = now.diff(purchaseDate).shiftToAll();
const warrantyLeft = warranty.minus(age).shiftToAll();

console.log({
  warrantyLeft: warrantyLeft.toString(),
});
```
You can run it on https://npm.runkit.com/luxon and see the unexpected `{warrantyLeft: "P99Y11M3.3481250000000005W"}` in the output.

```ts
const { DateTime, Duration } = require("luxon@3.3.0");

const purchaseDate = DateTime.fromObject(
  { day: 1, month: 1, year: 2020 },
  { zone: "UTC", conversionAccuracy: "longterm" }
);
const warranty = Duration.fromObject({ years: 100 }, { zone: "UTC", conversionAccuracy: "longterm" });
const now = DateTime.fromObject(
  { day: 8, month: 1, year: 2020 },
  { zone: "UTC", conversionAccuracy: "longterm" }
);
const age = now.diff(purchaseDate).shiftToAll();
const warrantyLeft = warranty.minus(age).shiftToAll();

console.log({
  warrantyLeft: warrantyLeft.toString(),
});
```
You can run it on https://npm.runkit.com/luxon and see the expected `{warrantyLeft: "P99Y11M3W2DT10H29M6S"}` in the output.

- OS: WSL2 in Windows
- Browser N/A
- Luxon version 3.4.2
- Your timezone Australia/Sydney

**Additional context**
Might be loosely related to https://github.com/moment/luxon/issues/1482 ?

diesieben07:
This fixes #1496 by introducing an additional conversion step in `Duration#normalize` / `normalizeValues`.
This step converts any fractions in larger units into lower units, if possible. 
For example: `{ years: 2.5, days: 0, hours: 0 }` becomes `{ years: 2, days: 182, hours: 12 }` (assuming casual conversions).

This also affects `shiftTo` and `shiftToAll`. I have added tests for all three methods to test this behavior.

Note: This alters the behavior of `normalize`, because previously `normalize` would not do this. `{years: 2.5, days: 0}` would stay as-is. I would argue that that is a bug however, which is fixed by this pull request. `shiftTo` (and by extension `shiftToAll`) _did_ do this, prior to calling `normalize`, however I would argue that this is something that should be handled by `normalize`.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -734,7 +734,7 @@ export default class Duration {
    * @example Duration.fromObject({ years: 2, days: 5000 }).normalize().toObject() //=> { years: 15, days: 255 }
    * @example Duration.fromObject({ days: 5000 }).normalize().toObject() //=> { days: 5000 }
    * @example Duration.fromObject({ hours: 12, minutes: -45 }).normalize().toObject() //=> { hours: 11, minutes: 15 }
-   * @example Duration.fromObject({ years: 2.5, days: 0 })
+   * @example Duration.fromObject({ years: 2.5, days: 0, hours: 0 }).normalize().toObject() //=> { years: 2, days: 182, hours: 12 }
    * @return {Duration}
    */
   normalize() {
--- a/src/duration.js
+++ b/src/duration.js
@@ -141,49 +141,51 @@ function normalizeValues(matrix, vals) {
   // the logic below assumes the overall value of the duration is positive
   // if this is not the case, factor is used to make it so
   const factor = durationToMillis(matrix, vals) < 0 ? -1 : 1;
-  const presentUnits = reverseUnits.filter((u) => !isUndefined(vals[u]));
 
-  presentUnits.reduce((previous, current, currentIndex) => {
-    if (previous) {
-      const previousVal = vals[previous] * factor;
-      const conv = matrix[current][previous];
+  orderedUnits.reduceRight((previous, current) => {
+    if (!isUndefined(vals[current])) {
+      if (previous) {
+        const previousVal = vals[previous] * factor;
+        const conv = matrix[current][previous];
 
-      // if (previousVal < 0):
-      // lower order unit is negative (e.g. { years: 2, days: -2 })
-      // normalize this by reducing the higher order unit by the appropriate amount
-      // and increasing the lower order unit
-      // this can never make the higher order unit negative, because this function only operates
-      // on positive durations, so the amount of time represented by the lower order unit cannot
-      // be larger than the higher order unit
-      // else:
-      // lower order unit is positive (e.g. { years: 2, days: 450 } or { years: -2, days: 450 })
-      // in this case we attempt to convert as much as possible from the lower order unit into
-      // the higher order one
-      //
-      // Math.floor takes care of both of these cases, rounding away from 0
-      // if previousVal < 0 it makes the absolute value larger
-      // if previousVal >= it makes the absolute value smaller
-      const rollUp = Math.floor(previousVal / conv);
-      vals[current] += rollUp * factor;
-      vals[previous] -= rollUp * conv * factor;
-      // try to convert any decimals into units smaller than "previous"
-      // this happens especially with conversionAccuracy: longterm
-      // for example for { years: 2, days: 450, seconds: 0 } we get { years: 3, days: 84.7575, seconds: 0 } at first
-      // we want to convert the fractional days into the lower order unit seconds
-      for (let i = currentIndex - 2; i >= 0; i--) {
-        let fraction = vals[previous] % 1;
-        if (fraction === 0) break;
-        const lowerOrderUnit = presentUnits[i];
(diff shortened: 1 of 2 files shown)


icambron:
Moving fractional parts of bigger units into smaller units is totally something normalize() should do

jayaddison:
A couple of questions about this:

* If the fraction of the larger unit would result in a fractional (non-decimal) number of the smaller units, will the smaller unit still be adjusted?
* Could this make `normalize` non-idempotent?  (as in, `normalize(normalize(n)) != normalize(n)`)
* Maybe a question about `normalize` more generally (I don't see it in the documentation currently) - is there a way to configure the conversion matrices used during normalization?  Years and months in particular seem like they can be ambiguous in duration.

diesieben07:
@jayaddison 
> If the fraction of the larger unit would result in a fractional (non-decimal) number of the smaller units, will the smaller unit still be adjusted?

Yes, fractions will be moved down as much as possible. Or, worded differently, only the smallest unit should have fractions after normalization.

> Could this make normalize non-idempotent? (as in, normalize(normalize(n)) != normalize(n))

Not as far as I can tell, because the 2nd call would not find any fractions to move to smaller units. Do you have a situation in mind where this could happen?

> Maybe a question about normalize more generally (I don't see it in the documentation currently) - is there a way to configure the conversion matrices used during normalization? Years and months in particular seem like they can be ambiguous in duration.

The matrix can be configured on the Duration itself and is used for more than just normalization. You can find out more in [the documentation](https://moment.github.io/luxon/#/math?id=casual-vs-longterm-conversion-accuracy).

jayaddison:
Thanks very much @diesieben07 - that's reassuring, and resolves all my questions.  Also my apologies - I didn't find that section of the documentation.  It's an enjoyable read!

The concern I had about a non-idempotent case was not a realistic one as long as the normalization is a strictly a single-pass greatest-to-smallest: I had wondered about, for example, fractional years being shifted down to `days`, but that resulting in a value that is above a `month`-or-`week` threshold, and therefore a subsequent call translating those `days` 'up' into a larger unit.  It sounds like that cannot happen.

jayaddison:
Hold on, maybe my concern returns due to this in the API docs:

> excessive values for lower-order units are converted to higher-order units (if possible, see first and second example)

As usual I should read and test more before commenting.  I will take more time to look into this soon.

So conversion 'up' could occur?

diesieben07:
Yes, conversion up will occur, but only for parts of lower order units that wholly fit into higher order units. 
For example `{ years: 0, months: 13 }` will become `{ years: 1, months: 1 }`.

This cannot be in conflict with the first part, because we only convert fractional parts of the higher order units downwards. This means that those parts will never be considered for upward conversion, because they would not fit into an integer part of the higher unit.

Example:
In `{years: 0.5, months: 0}` we convert the 0.5 years into 6 months. Those 6 months will never be considered for upwards conversion, because they would not turn into a whole year.

jayaddison:
@diesieben07 Ok; experimenting with this has reassured me but I did find one more caveat:

* The good: I now understand that both 'upwards' and 'downwards' normalization do not intoduce any new unit fields into an existing `Duration` object.  So even if a duration has `years: 1, days: 7`, no `weeks` units would be introduced.  That's a helpful constraint in this context that makes reasoning about this change easier.

* The caveat: if the 'downwards' normalization of a fractional value into the following unit causes the smaller unit to rollover the threshold at which it becomes greater-than-one of the larger unit, then normalization (in terms of the `.values`) is non-idempotent.  It's probably easier to explain with an example:

```
> Duration = require('luxon').Duration;
> Duration.fromISO('P2.5Y11M').normalize().values
{ years: 2, months: 17 }
> Duration.fromISO('P2.5Y11M').normalize().normalize().values
{ years: 3, months: 5 }
> Duration.fromISO('P2.5Y11M').normalize().equals(Duration.fromISO('P2.5Y11M').normalize())
true
> Duration.fromISO('P2.5Y11M').normalize().equals(Duration.fromISO('P2.5Y11M').normalize().normalize())
false
```

I understand that the `equals` check requires the same units _and_ values -- it's not a duration-equivalence check -- even so that could be confusing in some situations.

PR #1499, opened 2023-09-05
Added some JS doc for time zones

ptandler:
Had to look at the sources to get an idea of what the IANA zones and the UTC zones are really doing. Maybe this helps.


ptandler:
@icambron I added some documentation that I was missing when working with Luxon. Might be helpful for others as well.

I just amended the commit such that it uses my github email address and the CLA check now succeeded.

icambron:
Thanks!

Issue #849, opened 2021-01-07
IE - hackyoffset broken if computer date format does not include "/"


fromkeith:
**Describe the bug**
`hackyoffset` regex `/(\d+)\/(\d+)\/(\d+),? (\d+):(\d+):(\d+)/` expects the delimiter to be `/`. If a computer date time is set to a different delimiter (eg `-`) `hackyoffset` will break. This results in any use of `hackyoffset` to fail and throw an error. Explicitly, the regex will `exec` and result in `null`. 

**To Reproduce**
Please share a minimal code example that triggers the problem:

_Setup:_
1. On your computer (Windows 10), change the date time format:
     1. right click the clock on the bottom left.
     2. adjust date/time
     3. on the left side menu select "region"
     4. select "change date formats" on the bottom of the page
     5. select a short date like "2017-04-05"
     6. close the window.
6. open up test script in IE11. See the error!

```html
<!-- poly fills -->
<script src="https://cdn.polyfill.io/v2/polyfill.js?features=default,String.prototype.repeat,Array.prototype.find,Array.prototype.findIndex,Math.trunc,Math.sign"></script>
    <script src="https://unpkg.com/date-time-format-timezone@latest/build/browserified/date-time-format-timezone-complete-min.js"></script>
<!-- luxon -->
<script src="https://cdn.jsdelivr.net/npm/luxon@1.25.0/build/global/luxon.min.js"></script>
<!-- breaking code -->
<script>
console.log(luxon.DateTime.fromSeconds(0, {zone: 'America/Vancouver'}).toFormat('yyyy'))
</script>
```

**Actual vs Expected behavior**

Actual. Error thrown: `Unable to get property '1' of undefined or null reference`. This is due to the regex in hackyoffset returning `null`.

Expected log: 1969

**Desktop (please complete the following information):**

- OS: [e.g. iOS] Windows 10
- Browser [e.g. Chrome 84, safari 14.0] IE 11
- Luxon version [e.g. 1.25.0] 1.25.0
- My timezone: America/Toronto
- Computer default language: English (United States)

**Additional context**

Proposed fix: 

```javascript
function hackyOffset(dtf, date) {
  const formatted = dtf.format(date).replace(/\u200E/g, ""),
    parsed = /(\d+)[\/-](\d+)[\/-](\d+),? (\d+):(\d+):(\d+)/.exec(formatted),
    [, fMonth, fDay, fYear, fHour, fMinute, fSecond] = parsed;
[...]

PR #1503, opened 2023-09-08
Improve IANAZone.offset performance

mohd-akram:
This is more than 2.5x faster in my testing. Example benchmark script:

```javascript
import { setTimeout } from 'timers/promises';

import { DateTime } from './src/luxon.js';

// Warm up
for (let i = 0; i < 10; i++) {
  DateTime.fromJSDate(new Date(), { zone: 'America/New_York' });
}

await setTimeout(200);

console.time('run');
for (let i = 0; i < 10000; i++) {
  DateTime.fromJSDate(new Date(), { zone: 'America/New_York' });
}
console.timeEnd('run');
```

### DIFF
--- a/src/zones/IANAZone.js
+++ b/src/zones/IANAZone.js
@@ -1,9 +1,23 @@
 import { formatOffset, parseZoneInfo, isUndefined, objToLocalTS } from "../impl/util.js";
 import Zone from "../zone.js";
 
-const dtfCache = new Map();
-function makeDTF(zoneName) {
-  let dtf = dtfCache.get(zoneName);
+const directOffsetDTFCache = new Map();
+function makeDirectOffsetDTF(zoneName) {
+  let dtf = directOffsetDTFCache.get(zoneName);
+  if (dtf === undefined) {
+    dtf = new Intl.DateTimeFormat("en-US", {
+      timeZone: zoneName,
+      timeZoneName: "longOffset",
+      year: "numeric",
+    });
+    directOffsetDTFCache.set(zoneName, dtf);
+  }
+  return dtf;
+}
+
+const calculatedOffsetDTFCache = new Map();
+function makeCalculatedOffsetDTF(zoneName) {
+  let dtf = calculatedOffsetDTFCache.get(zoneName);
   if (dtf === undefined) {
     dtf = new Intl.DateTimeFormat("en-US", {
       hour12: false,
@@ -16,7 +30,7 @@ function makeDTF(zoneName) {
       second: "2-digit",
       era: "short",
     });
-    dtfCache.set(zoneName, dtf);
+    calculatedOffsetDTFCache.set(zoneName, dtf);
   }
   return dtf;
 }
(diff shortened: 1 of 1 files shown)


diesieben07:
The problem with this is that the output format of `longOffset` is not described in the specification, so there is no guarantee that it will work like this in all engines or even that it will keep working in the same way.

mohd-akram:
@diesieben07 This problem exists for `hackyOffset` and `partsOffset` as well. Luxon generally makes certain assumptions about the engine, such as it having `Intl.DateTimeFormat`, an `en-US` locale, and a format that looks a certain way. It is recommended for engines to use the CLDR data, and in the case of long offset formatting it is [fairly rigid](https://cldr.unicode.org/development/development-process/design-proposals/localized-gmt-format).

diesieben07:
`hackyOffset` is only used if `formatToParts` is not supported, which is pretty much only in ancient runtimes. I don't see how `partsOffset` is susceptible to the same issue, as it looks at each part individually and only parses numbers.
At the very least we should check that the expected format is returned by `longOffset` and only set `hasLongOffset` if that is so.

mohd-akram:
> `hackyOffset` is only used if `formatToParts` is not supported, which is pretty much only in ancient runtimes. I don't see how `partsOffset` is susceptible to the same issue, as it looks at each part individually and only parses numbers.

`partsOffset` is certainly less susceptible, perhaps the only part where it might fail is the `BC` check.

> At the very least we should check that the expected format is returned by `longOffset` and only set `hasLongOffset` if that is so.

I've added this. I also extracted the offset calculations out of the method, and made each function get its own DTF since they rely on different formats. The check is expensive, that's why I had to do it inside the `offset` method as otherwise importing the library is slower (the `Intl.DateTimeFormat` constructor is painfully slow).

schleyfox:
I found this issue while investigating a performance issue in offset that impacted a use case of parsing a few hundred thousand dates at a time.

I noticed that `hackyOffset` was substantially faster than `partsOffset` (about 2.5x on node 18). I wrote a [benchmark to compare](https://measurethat.net/Benchmarks/Show/28510/0/intldatetimeformat-formattoparts-vs-format-then-regex) and got even more dramatic results in browser


Test case name | Result
-- | --
hackyOffset | hackyOffset x 718,239 ops/sec ±0.67% (68 runs sampled)
partsOffset | partsOffset x 208,390 ops/sec ±0.21% (69 runs sampled)
dtf.formatToParts | dtf.formatToParts x 298,653 ops/sec ±0.52% (68 runs sampled)
dtf.format | dtf.format x 822,318 ops/sec ±0.18% (60 runs sampled)


It seems like there might be [good reasons](https://github.com/moment/luxon/issues/849) not to enable this in the browser, but I am very tempted to use this approach in node where I control the environment...

mohd-akram:
Any update on this? I've been using it in production since and it seems to be working well. Note that this is even faster than `hackyOffset` (1.4x as fast in my testing).

PR #1507, opened 2023-09-15
Fix docstring example reference

pouwerkerk:
There appears to be a very small typo in the docstrings for `Duration.normalize()`. The docstring mentions negative values, but points to an example about excessive values, whereas the _third_ example appears to exemplify the point made in the docstring.

Screenshot 2023-09-15 at 9 38 58 AM


icambron:
thanks

PR #1508, opened 2023-09-16
perf(partsOffset): allocate fixed size array

amilajack:


diesieben07:
An even better way to do this would be to just use `Array#map` on `formatted`. This way we avoid having a "holey" or sparse array (i.e. an array with holes in it), which JS engines are less able to optimize.

Issue #1506, opened 2023-09-14
toLocaleParts offby 1 Error on dates prior to "1884-01-01"


UpsidePotential:
**Describe the bug**
.toLocaleParts returns a date offby 1 if used on a date prior to 1884-01-01.  formatToParts works correctly.

**To Reproduce**
```
const { DateTime } = require("luxon")
const timeString = "1800-01-01T00:00:00.0000000";

{
    let time = DateTime.fromISO(timeString)
    time.setLocale('en-US')
    const parts = time.toLocaleParts({ year: 'numeric', month: '2-digit', day: '2-digit' })

    const partValues = parts.map((p) => p.value);
    console.log(partValues);
}

{
    const date = new Date(timeString);
    const options = { year: 'numeric', month: '2-digit', day: '2-digit' };
    const dateTimeFormat = new Intl.DateTimeFormat('en-US', options);

    const parts = dateTimeFormat.formatToParts(date);
    const partValues = parts.map((p) => p.value);
    console.log(partValues);
}
```

```
["12", "/", "31", "/", "1799"]
["01", "/", "01", "/", "1800"]
```

**Actual vs Expected behavior**
Expect toLocaleParts to return the same values as formatToParts.

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser Chrome 116.0.5845.188
- Luxon version 3.4.3
- Your timezone "America/Chicago"

icambron:
Thanks for the report. This issue's root isn't in formatting, it's converting to the native date. The problem with the date is that it has an offset with a seconds component. You can see that here:

```js
d  = new Date("1800-01-01T00:00:00.0000000")   // => 1800-01-01T04:56:02.000Z 
```

Those 2 seconds are there because the offset is 296 minutes and 2 seconds. The problem is that JS lies about this:

```js
d.getTimezoneOffset()   //=> 296
```

The 2 seconds are gone. That's because the JS date API simply doesn't expose seconds. Luxon needs to convert the date into a native date to pass it to the Intl API, so it adds the current offset (296) to UTC time and calls `new Date()` on it. Unfortunately, that offset is 2 seconds short and leaves the date as 23:59:58 the day before. Then we feed the this native date into Intl and it happily prints out the wrong date.

[...]

Issue #1399, opened 2023-03-17
Argentina timezone is wrong before the day 1920/05/01


augustogalan7:
On 1 May 1920 Argentina's time zone changed from UTC-4:16 to UTC-4, when you apply the toLocaleString method on that date, the result is one day earlier.

When you use the browser's native functions you get
new Date(1920,04,02).toLocaleTimeString() ---> '0:00:00'.
new Date(1920,04,01).toLocaleTimeString() ---> '0:16:48'.

Can be reproduced in code pen:
var DateTime = luxon.DateTime;
console.log("Date is incorrect expected on 1/5/1920 but found on 30/4/1920")
console.log(DateTime.local(1920, 05, 01).toLocaleString(DateTime.DATE_SHORT))

link: https://codepen.io/Vakhnina/pen/RMXRMx

icambron:
There are a couple things going on here.

The first is that the offset isn't just -4:16; it's -4:16:48. The problem for Luxon is that JS doesn't report the seconds:

```js
new Date(1920, 3, 30).getTimezoneOffset()   //=> 256 
```

This throws off Luxon's computations by exactly those seconds. To take an arbitrary date before the switch:

```js
DateTime.local(1920, 2, 1).toLocaleString(DateTime.DATETIME_FULL_WITH_SECONDS) //=> 'January 31, 1920 at 11:59:12 PM GMT-4:16:48'  
```

Notice Luxon is short those exact 48 seconds, which unfortunately puts the date into the previous day. For more, see #401.

There's more going on here though, because 1920-05-01 is after the switch. I think what's suppose to happen is that at midnight, the clocks advanced from midnight to 00:16:48. This means that local times like 00:00:00 or, say, 00:10:15 never actually happened. In theory, if you try to build a local time with those "hole times" you should get '0:16:48' for them, but V8 doesn't seem to handle it that way:

```js
> new Date(1920, 3, 30, 23, 59, 59).toLocaleString()                                                                                                                                
'4/30/1920, 11:59:59 PM'                                                                                                                                                                                                                                                                                                                  
> new Date(1920, 4, 1, 0, 0).toLocaleString()                                                                                                                                       
'5/1/1920, 12:16:48 AM'                                                                                                                                                             
> new Date(1920, 4, 1, 0, 15).toLocaleString()                                                                                                                                      
'5/1/1920, 12:31:48 AM'                                                                                                                                                             
[...]

Issue #684, opened 2020-04-02
GMT+0119 timezone handling


entioentio:
Between dates 30 June and 1 July of 1937 there's timezone change from GMT+0019 to GMT+0120. When using `fromJSDate`, the `o` property of DateTime object is a float, but only in FF & IE, not in chrome (didn't test other browsers).
I'm sure it's browser dependant and can be worked around in my front-end code, but the question is if shouldn't it be taken into consideration in the lib's code.

See the screenshot from FF:
js `date`,
`.fromJSDate`
`.toFormat`

![image](https://user-images.githubusercontent.com/10046841/78236316-17403d80-74da-11ea-8f9e-cd4ed074c04d.png)

icambron:
I'm actually a little confused by the results there. It looks to me like it shifted from +1:19:32 to +1:20. My guess is that is that Chrome simply doesn't know about the 32 seconds, which unfortunately means Luxon doesn't understand it either. What exact time zone is this? (e.g. "Europe/Warsaw" or similar) I wasn't able to immediately find this historical data.

GillesDebunne:
The offset was `19m 32.13s` until July 1st 1937 (see [UMT+00:20](https://en.wikipedia.org/wiki/UTC%2B00:20)) in the Netherlands for timezone _'Europe/Amsterdam'_.

This value is rounded to `19m` in Chrome, and to `19m 32s` in FF which results in the `79.5333333` of your screenshot.

#755 rounds the displayed offset to the minute, which will now display `+01:19` in all browsers. However, since the offset is different between browsers, this may lead to discrepancies...

An option could be to always round the offset (expressed in minutes) in DateTime constructors.

PR #1509, opened 2023-09-16
use ISO string when converting system zone to JS date

icambron:
I'm of two minds about whether this is a good idea or not. This is to fix issues where historic zones have offsets that aren't whole minutes. Examples:

* https://github.com/moment/luxon/issues/1506
* https://github.com/moment/luxon/issues/1399
* https://github.com/moment/luxon/issues/684

I'm sure there are others. Anyway, if it's a SystemZone date, we are using `getTimezoneOffset` (which provides whole minutes) to compute the epoch milliseconds from the input string, which will be off by however many seconds in the offset. When we convert we convert that to a JS date (for example, to format it), it will then be the wrong date. This fixes that by providing the ISO string directly to the JS Date, which will parse it using the more accurate offset the native date has access to internally.

The problem with the fix is that it may lead to other inconsistencies. The epoch millis in the DateTime will be different from the epoch millis in the native date. It's not clear to me whether this is making things better or merely making them _weirder_.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -47,6 +47,7 @@ import {
   InvalidDateTimeError,
 } from "./errors.js";
 import Invalid from "./impl/invalid.js";
+import SystemZone from "./zones/systemZone.js";
 
 const INVALID = "Invalid DateTime";
 const MAX_DATE = 8.64e15;
@@ -1887,7 +1888,13 @@ export default class DateTime {
    * @return {Date}
    */
   toJSDate() {
-    return new Date(this.isValid ? this.ts : NaN);
+    if (!this.isValid) {
+      return new Date(NaN);
+    } else if (this.zone == SystemZone.instance) {
+      return new Date(this.toISO({ includeOffset: false }));
+    } else {
+      return new Date(this.ts);
+    }
   }
 
   // COMPARE
--- a/test/datetime/transform.test.js
+++ b/test/datetime/transform.test.js
@@ -63,6 +63,12 @@ test("DateTime#toJSDate() returns a native Date equivalent", () => {
   expect(js.getTime()).toBe(dt.toMillis());
 });
 
+test("DateTime#toJSDate() works for historic dates with weird offsets", () => {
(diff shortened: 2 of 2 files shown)


diesieben07:
`IANAZone#offset` already does weird things to get the offset: It creates a `DateTimeFormat` with its timezone and then does trickery to extract the offset. This works fine (although we should probably not use floating point fractions here due to precision):
```js
const ts = -5364644638000; // 1800-01-01T00:00:00.0000000 in America/New_York
console.log(IANAZone.create("America/New_York").offset(ts)); // outputs -296.03333333333336
```

We can use the same trickery in `SystemZone`, except not give a zone name to `Intl.DateTimeFormat`, so it uses the system zone.
The only issue with this is that either we have to recreate the DTF every time or cache it, in which case it would report the wrong value when the system's zone changes. But I think we have that same issue in other places already.
Additionally we might be able to find a way to know whether we can get away with just calling `Date#getTimezoneOffset` vs. going the Intl route.

icambron:
Yeah, the IANA zone handles this just fine (though we should probably change the offset interface to return seconds). I have been loathe to port this logic to the system zone just because every Intl operation is so slow. The vast majority of real world uses are not on historic time zones, so this would be a poor tradeoff.

Yes, perhaps a heuristic on when to use which would work. Old system datetimes would be slower and new ones would be fast. We'd just have to pick the cutoff carefully.

diesieben07:
For the system zone we can get away without using Intl, because `Date` uses the system zone:
```js
const date = new Date(-5364644638000);
const fakeUTC = Date.UTC(
  date.getFullYear(), date.getMonth(), date.getDate(),
  date.getHours(), date.getMinutes(), date.getSeconds(), date.getMilliseconds()
);
const offset = (date.valueOf() - fakeUTC) / 1000;
console.log(date.getTimezoneOffset() * 60, offset); // 17760, 17762
```

icambron:
That's a good point

PR #1523, opened 2023-10-08
Add next/previous weekday enhancement

### LINKED ISSUE #1517: Inconsistent set weekday behavior
**Describe the bug**
DateTime.set({ weekday: X }) sometimes returns a date in the past, and sometimes in the future.
**To Reproduce**
```js
luxon.DateTime.fromISO('2023-10-01T20:24:00Z').set({ weekday: 3 }).toISO()
>'2023-09-27T16:24:00.000-04:00'
luxon.DateTime.fromISO('2023-10-02T20:24:00Z').set({ weekday: 3 }).toISO()
>'2023-10-04T16:24:00.000-04:00'
```
All values return a future date

- Browser [e.g. Chrome 84, safari 14.0]
- Luxon version [e.g. 1.25.0]
- Your timezone [e.g. "America/New_York"]

**Additional context**
I think in moment.js, setting a weekday will always return a future date

shashankbhat10:
I have a couple of questions regarding this.

1. What should be the expected behaviour when weekday argument is incorrect(eg: weekday is a string or the weeday value is out of valid range). Should the method be throwing an InvalidArgumentException?

2. Currently, the method will return the DateTime with the next/previous weekday. So, if the current weekday is Monday(1) and the methods are called with argument as 1, the method does not return the current date but the previous or next Monday. Is this the correct behaviour? 

resolves #1517

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -2039,6 +2039,34 @@ export default class DateTime {
   }
 
   /**
+   *
+   * @param {number} weekday - The number of the required weekday (1 is Monday and 7 is Sunday)
+   * @returns {DateTime}
+   * @example DateTime.fromISO('2023-10-06T09:08:34.123').nextWeekday(weekday = 1) //=> Next Monday
+   * @example DateTime.fromISO('2023-10-06T09:08:34.123').nextWeekday(weekday = 5) //=> Next Friday
+   */
+  nextWeekday(weekday) {
+    if (!this.isValid) return this;
+
+    let daysToAdd = (7 - this.weekday + weekday) % 7;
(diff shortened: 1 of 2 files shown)


thomassth:
@shashankbhat10 you'll need to authorize on the CLA for this PR

shashankbhat10:
@thomassth @diesieben07 I think I messed up while creating the PR as the commit is from a different account. Is it fine if I create a duplicate PR with commit from this ID (I have signed the CLA on this account).

shashankbhat10:
Closing PR (due to CLA conflict). 
Handled issue in #1538

PR #1524, opened 2023-10-10
Fix Interval.splitAt datetime sorting

joshuajaco:
`Interval#splitAt` is incorrectly sorting the input `DateTime`s by ISO strings (`Array.sort` implicitly calls `toString`), which leads to invalid intervals when trying to split at DST shifted hours.
This PR changes the sorting to use the underlying epoch milliseconds instead.

### DIFF
--- a/src/interval.js
+++ b/src/interval.js
@@ -312,7 +312,7 @@ export default class Interval {
     const sorted = dateTimes
         .map(friendlyDateTime)
         .filter((d) => this.contains(d))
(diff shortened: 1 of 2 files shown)


joshuajaco:
@icambron anything I can do to get this merged?
This is currently preventing us from using `splitAt`

diesieben07:
Thank you for this bug fix, @joshuajaco!

Issue #318, opened 2018-08-15
Module `util` does not exist in React Native


ntelkedzhiev:
I'm trying to build my react-native app (RN: 0.56.0) and I'm getting the following error: 

`error: bundling failed: Error: Unable to resolve module util from ../node_modules/luxon/build/cjs-browser/luxon.js: Module util does not exist in the Haste module map or in these directories:
  /../../../node_modules`

Do I have to explicitly add `util` as a package?

icambron:
Hmm, no. I'm not even sure what that means.

ivosabev:
@icambron maybe here: https://github.com/moment/luxon/blob/2b65af4a58b91039934628961b468ef803f939e6/src/impl/util.js#L207

tuananh:
`util` is a core Node module right?

ivosabev:
Not in `react-native`, it has to be installed additionally from https://www.npmjs.com/package/utils

tuananh:
@ivosabev that's `utils` with `s`. `util` is a core Node module iirc

ivosabev:
My bad, wrong link. I meant https://www.npmjs.com/package/util

icambron:
Hmm, yeah. Is there a way in react-native to check if the `require` will fail? The problem is that it can't be handled in the catch.

ivosabev:
No there is no way. RN will bundle all packages and user code before running the application. The error actually comes at the bundling stage.

I think it is best to add a message that the `util` polyfill is needed if you want to use Luxon on React Native, just like you have for the Intl support or use an alternative to `util.inspect`?

icambron:
I'd prefer to remove the the dependency then. It's not worth causing so much pain.

icambron:
I removed `inspect` altogether in 55ac17e6c0ed936bb098d77f933a5afbe24c787b, will release it soon

PR #1526, opened 2023-10-10
Add custom inspect for node without depedency

pedropedruzzi:
Adds back the custom inspection functions for node.js, first added by #281 then removed by #318 and [55ac17e]. This change avoids the default node's object inspection listing all fields recursively. This time there are no node imports or any run-time dependency. We simply expose the inspect functions as symbol-named attribute `Symbol.for('nodejs.util.inspect.custom')` as described in the [docs](https://nodejs.org/api/util.html#utilinspectcustom).

Note: I've simplified the formats from the original addition to make it single line and simpler.

## Testing done
* npm run build
* npm run format-check
* ./scripts/test
* Manual
```
const { DateTime, Duration, Interval } = require('./build/cjs-browser/luxon')
Duration.fromObject({ seconds: 10 })
Interval.fromDateTimes(DateTime.now().minus({ days: 1 }), DateTime.now())
DateTime.now()

// output:
Duration { values: {"seconds":10} }
Interval { start: 2023-10-09T16:20:51.017-03:00, end: 2023-10-10T16:20:51.019-03:00 }
DateTime { ts: 2023-10-10T16:20:59.623-03:00, zone: America/Sao_Paulo, locale: en-US }
```

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -608,6 +608,18 @@ export default class Duration {
   }
 
   /**
+   * Returns a string representation of this Duration appropriate for the REPL.
+   * @return {string}
+   */
+  [Symbol.for("nodejs.util.inspect.custom")]() {
+    if (this.isValid) {
+      return `Duration { values: ${JSON.stringify(this.values)} }`;
+    } else {
+      return `Duration { Invalid, reason: ${this.invalidReason} }`;
+    }
+  }
+
+  /**
    * Returns an milliseconds value of this Duration.
    * @return {number}
    */
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1815,6 +1815,18 @@ export default class DateTime {
   }
 
   /**
+   * Returns a string representation of this DateTime appropriate for the REPL.
+   * @return {string}
+   */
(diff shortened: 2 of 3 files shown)


src/duration.js
```diff
@@ -607,6 +607,18 @@ export default class Duration {
     return this.toISO();
   }
 
+  /**
+   * Returns a string representation of this Duration appropriate for the REPL.
+   * @return {string}
+   */
+  [Symbol.for("nodejs.util.inspect.custom")]() {
+    if (this.isValid) {
+      return `Duration { values: ${JSON.stringify(this.values)} }`;
```
pedropedruzzi, line 616: Author note: I've used JSON.stringify for simplificy but it makes Duration's format different from the others. If maintainers prefer, I could change to something like:

```suggestion
      return `Duration { ${Object.entires(this.values).map(([key, val]) => `${key}: ${val}`).join(', ')} }`;
```

icambron: I think the json stringify is probably right

icambron:
This looks good, thanks

PR #1534, opened 2023-11-02
Improve Duration#toHuman docblock

Renegade334:
The `Duration#toHuman` docblock is currently quite unclear as to where `listStyle` comes from, which hopefully this improves.


icambron:
Thanks!

PR #1538, opened 2023-11-09
Enhancement for next/previous weekday

shashankbhat10:
I have a couple of questions regarding this. @diesieben07 

1. What should be the expected behaviour when weekday argument is incorrect(eg: weekday is a string or the weeday value is out of valid range). Should the method be throwing an InvalidArgumentException?

2. Currently, the method will return the DateTime with the next/previous weekday. So, if the current weekday is Monday(1) and the methods are called with argument as 1, the method does not return the current date but the previous or next Monday. Is this the correct behaviour?

resolves https://github.com/moment/luxon/issues/1517

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -2164,6 +2164,34 @@ export default class DateTime {
   }
 
   /**
+   *
+   * @param {number} weekday - The number of the required weekday (1 is Monday and 7 is Sunday)
+   * @returns {DateTime}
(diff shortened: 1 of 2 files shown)


icambron:
Thanks for this contribution and sorry for the log delay in reviewing this. I think we'll pass on this. It's not that it isn't correct; it's that it's easy enough to do in "userspace". We have tried (with mixed success) to avoid feature creep like this.

PR #1546, opened 2023-11-28
Alias DateTime in documentation site

forgotPassword:
All examples use `DateTime`. The install guide even suggests doing the alias.

The first thing I tried was to copy the example on the landing page `DateTime.now().setZone('America/New_York').minus({weeks:1}).endOf('day').toISO();`
->
`Uncaught ReferenceError: DateTime is not defined at <anonymous>:1:1`


icambron:
thanks

PR #1548, opened 2023-12-03
Create SECURITY.md

Tonmoyhridhaan:


fishcharlie (changes requested):
This should be closed. Any `SECURITY.md` should come from the official maintainers. Additionally this is just importing the default template which doesn't work for this library (see the version section).

icambron:
Security fixes are very rare around here, and when they come up, we'll decide which versions they go into on an ad hoc basis

PR #1562, opened 2024-01-06
Return Invalid DateTime for using meridiem with 24-hour (#1549)

### LINKED ISSUE #1549: DateTime is not invalid like it should be when using hour 15 with 12-hour time token.
**Describe the bug**
Correct me if I'm wrong, but shouldn't this return an invalid DateTime?

```typescript
luxon.DateTime.fromFormat('11-08-2023 15:00 AM', 'MM-dd-yyyy hh:mm a').toString()
// '2023-11-08T15:00:00.000-06:00'
```

**To Reproduce**
Just go to the Luxon docs and paste the example above into the console.
image

The date string supplied is using hour 15 with a meridian, and the format string is expecting 12-hour time. Shouldn't luxon notice that hour 15 is out of bounds and return an Invalid object?

- OS: MacOS Ventura 13.1
- Browser: Chrome 118.0.5993.117
- Luxon version: 3.4.4
- Your timezone: America/Chicago

Matthematic:
This is a fix for the case where the input **'11-08-2023 15:00 AM'** is parsed and interpreted as **'11-08-2023 03:00 PM'**. Instead of this, it should be a Datetime.Invalid with a "unit out of range" reason similar to what happens with **'11-08-2023 25:00'**.

fixes #1549


icambron:
Sorry for the long delay in reviewing this.

While I agree this is a bug, it's a minor one, and the fix is unfortunate, because it's extra work on the hot path. I think if you can use the benchmarks not mearningfully shower that token parsing is no slower now, we can merge this.

PR #1564, opened 2024-01-09
extended jsdoc about duration and units

ptandler:
especially when using Duration#diff or Duration#shiftTo


ptandler:
@icambron After understanding a bug I had using Luxon, I thought it might be a good idea to add a bit to the documentation so others might prevent running into the same problem.

src/duration.js
```diff
@@ -725,6 +729,9 @@ export default class Duration {
 
   /**
    * Return the length of the duration in the specified unit.
+   *
+   * Note as this uses {@link Duration#shiftTo} internally, it can cause the unexpected effects as described there.
```
icambron, line 733: Instead, this should say something like:

> Uses {@link Duration#shiftTo} to convert to the request unit. See the documentation there for details.

The reason this is better is that "unexpected" depends on your expectations. We don't want a _warning_ message; we want a _help you understand_ message.

src/duration.js
```diff
@@ -770,7 +780,25 @@ export default class Duration {
 
   /**
    * Convert this Duration into its representation in a different set of units.
-   * @example Duration.fromObject({ hours: 1, seconds: 30 }).shiftTo('minutes', 'milliseconds').toObject() //=> { minutes: 60, milliseconds: 30000 }
+   *
+   * Note that using shiftTo() can sometimes have unexpected results when using units that don't refer to a
```
icambron, line 784: Here too. Now like 95% of the doc for this is a troubleshooting guide for conversionAccuracy. I think here we just say:

> Converting between units is subtle because units don't always have constant length. Please see https://moment.github.io/luxon/#/math?id=casual-vs-longterm-conversion-accuracy

icambron (changes requested):
I really appreciate the effort you put into this. I've asked for some changes, mostly because the doc strings can't really be troubleshooting guides, but can be links out to detailed explanations in the other docs.

PR #1575, opened 2024-01-22
Validate time zone in quickDT prior to guessing offset

schleyfox:

We validate the zone in the constructor anyway, but the tests missed a branch where we attempt to use an invalid zone to compute an offset, which will result in an exception. Zone#isValid is calculated at creation time (and checked in DateTime constructor already) so this will not degrade performance in the happy case.

We discovered this when shifting callsites to use `DateTime.local` from things that previously called through `fromObject`.


icambron:
Looks good, thanks!

PR #1579, opened 2024-01-22
Perf: Cache ts offset guesses for quickDT

schleyfox:

_This includes the commit from https://github.com/moment/luxon/pull/1574 to establish the benchmark_

Previously we guessed by calling zone.offset(tsNow) on every
construction of a date from date parts. This caches that result instead,
saving a call to offset. The tests verify that this works across DST boundaries when the cache holds an offset for the previous DST state from now.

For processes that live across a DST boundary and mostly handle current date times, this will mean the guess is wrong and we'll have to re-guess. Since we're caching one call to offset, this should be performance neutral to current behavior. Otherwise, we save a call.

Benchmark Comparison (`name | before | after | after/before`):
```
DateTime.local with numbers and zone | 50,913 ±0.18% | 66,168 ±0.25% | 1.3x
```

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -390,14 +390,14 @@ function normalizeUnitWithLocalWeeks(unit) {
 // higher-order units from tsNow (as we do in fromObject, this requires that
 // offset is calculated from tsNow).
 function guessOffsetForZone(zone) {
-  if (!DateTime._zoneOffsetGuessCache[zone]) {
-    if (DateTime._zoneOffsetTs === undefined) {
-      DateTime._zoneOffsetTs = Settings.now();
+  if (!zoneOffsetGuessCache[zone]) {
+    if (zoneOffsetTs === undefined) {
+      zoneOffsetTs = Settings.now();
     }
 
-    DateTime._zoneOffsetGuessCache[zone] = zone.offset(DateTime._zoneOffsetTs);
+    zoneOffsetGuessCache[zone] = zone.offset(zoneOffsetTs);
   }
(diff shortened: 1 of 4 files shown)


src/datetime.js
```diff
@@ -369,13 +369,42 @@ function normalizeUnitWithLocalWeeks(unit) {
...
+// in the same offset (we could set the timestamp statically, instead). We use a
+// single timestamp for all zones to make things a bit more predictable.
+//
+// This is safe for quickDT (used by local() and utc()) because we don't fill in
+// higher-order units from tsNow (as we do in fromObject, this requires that
+// offset is calculated from tsNow).
+function guessOffsetForZone(zone) {
+  if (!DateTime._zoneOffsetGuessCache[zone]) {
+    if (DateTime._zoneOffsetTs === undefined) {
+      DateTime._zoneOffsetTs = Settings.now();
```
icambron, line 394: Should support clearing the cache [here](https://github.com/moment/luxon/blob/3125686af82d9a25c7267a1cf1eb838a3d41144f/src/settings.js#L171)

icambron (changes requested):

icambron:
Looks pretty good. Just requested a few changes

schleyfox:
Changed, thanks!

icambron:
@schleyfox if you fix the conflicts, I'll merge this

icambron (approved):

schleyfox:
@icambron conflicts fixed. thanks.

PR #1592, opened 2024-02-02
Use getPrototypeOf instead of __proto__

kapouer:
Debian checks what happens with running `node --disable-proto=throw`.
Because __proto__ brings so many trivial-to-exploit security issues, we might disable-proto by default.
Here it's just used in the tests, nonetheless, it's an old js feature that might just be phased out, hopefully.


icambron:
Sure, thanks

PR #1613, opened 2024-03-29
docs: until returns Inverval or DateTime conditionally

h4l-yup:
DateTime.until returns Interval or DateTime conditionally.

It is already on the [type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ec2fdb5dc09bcd3a871a46212c5f611b6cdd3a1c/types/luxon/src/datetime.d.ts#L1527) but luxon documentation is wrong with it.


icambron:
Thanks!

PR #1615, opened 2024-03-31
Update DateTime.equals() to cover the case when the param is not an instance of DateTime.

wangchj:
Update DateTime.equals() to cover the case when the param is not an instance of DateTime.


wangchj:
This org is extremely inefficient. This PR is very simple. Yes or no

PR #1627, opened 2024-05-15
Extend Duration documentation to document different behaviour when created out of factory methods

KlausNie:
* include note relevant methods
* include reference to documentation

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -864,6 +864,11 @@ export default class Duration {
 
   /**
    * Get the years.
+   *
+   * Note: Durations created out of Duration.from* (e.g. fromObject) factories will always yield 0.
+   * This will give you the correct difference in years, if the Duration was created out of a diff operation.
+   *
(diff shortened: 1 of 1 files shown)


src/duration.js
```diff
@@ -864,6 +864,11 @@ export default class Duration {
 
   /**
    * Get the years.
+   *
+   * Note: Durations created out of Duration.from* (e.g. fromObject) factories will always yield 0.
+   * This will give you the correct difference in years, if the Duration was created out of a diff operation.
```
diesieben07, line 869: This is factually incorrect. `Duration.fromObject({ years: 2 }).years` is `2`, not `0`.
I am not sure what you're referring to by "correct difference in years" here.

diesieben07 (changes requested):
Unfortunately the documentation you have added is factually not correct.
I also don't think we should be copy pasting the same text at all the accessors.

Issue #1625, opened 2024-05-14
DateTime fromFormat() is considers non-valid 24-hr formats as valid


eh1160:
**Describe the bug**
DateTime fromFormat() is considers non-valid 24-hr formats as valid

**To Reproduce**
DateTime.fromFormat('18:30 AM', 'h:mm a').isValid;  // = true

**Actual vs Expected behavior**
isValid should return false because 18 is not a valid "h" format

**Desktop (please complete the following information):**
- OS: MacOs Sonoma 14.2.1
- Browser Chrome Version 124.0.6367.119 (Official Build) (arm64)
- Luxon version ^3.4.4
- Your timezone "America/New_York"

Komock:
Have the same issue.
If I parse `16:00 PM` and print it with the same format, hours will be converted to 12-h format.
```
DateTime.fromFormat('16:00 PM', 'h:mm a').toFormat('h:mm a');
// '4:00 PM'
```

PR #1631, opened 2024-05-16
Fix/12 hour format validation

Sho-ki:
Fix issue: https://github.com/moment/luxon/issues/1625

## Description of the Bug
The `DateTime.fromFormat()` function incorrectly validates certain 24-hour formats as valid when they should be considered invalid. 
When using the 12-hour (h) format in the format parameter, the function does not properly invalidate times where the hour exceeds 12.

## Actual vs Expected Behavior
```javascript
DateTime.fromFormat('18:30 AM', 'h:mm a').isValid;
```
Actual: `isValid` returns true for `DateTime.fromFormat('18:30 AM', 'h:mm a')`.
Expected: `isValid` shouldn't be true because 18 is not a valid hour for the 12-hour (h) format.

## Fix Description
This PR addresses the issue by ensuring that when the 12-hour (h) format is used in the format parameter, any hour value greater than 12 is considered `invalid`, and isValid will throw error.

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -451,10 +451,13 @@ export class TokenParser {
     if (!this.isValid) {
       return { input, tokens: this.tokens, invalidReason: this.invalidReason };
     } else {
-      const [rawMatches, matches] = match(input, this.regex, this.handlers),
-        [result, zone, specificOffset] = matches
-          ? dateTimeFromMatches(matches)
-          : [null, null, undefined];
+      const [rawMatches, matches] = match(input, this.regex, this.handlers);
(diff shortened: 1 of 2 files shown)


diesieben07:
While this is a valid bugfix, I am reluctant to merge this in and release it as a minor update, because it is technically a breaking change. Normally that would be fine for a bug fix, but I can imagine many people intentionally or unintentionally relying on this behavior. Maybe we should add a `strictHours` option for backwards compatibility?

What do you think, @icambron?

jw642459986:
@diesieben07 are you suggesting something like this?
`DateTime.fromFormat('18:30 AM', 'h:mm a').isValid('strictHours': true);`

PR #1632, opened 2024-05-17
`refrenceDate` support for `fromObject`

angelaki:
Analogous to _date-fns_'s `parse` method, _luxon_ should support a `referenceDate`, too.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -745,8 +745,10 @@ export default class DateTime {
    * @param {string} [opts.locale='system\'s locale'] - a locale to set on the resulting DateTime instance
    * @param {string} opts.outputCalendar - the output calendar to set on the resulting DateTime instance
    * @param {string} opts.numberingSystem - the numbering system to set on the resulting DateTime instance
+   * @param {DateTime|Date|Object} opts.refrenceDate - the reference date to taken for missing parts
    * @example DateTime.fromObject({ year: 1982, month: 5, day: 25}).toISODate() //=> '1982-05-25'
    * @example DateTime.fromObject({ year: 1982 }).toISODate() //=> '1982-01-01'
+   * @example DateTime.fromObject({ year: 1982 }, { refrenceDate: { day: 10 } }).toISODate() //=> '1982-01-10'
    * @example DateTime.fromObject({ hour: 10, minute: 26, second: 6 }) //~> today at 10:26:06
    * @example DateTime.fromObject({ hour: 10, minute: 26, second: 6 }, { zone: 'utc' }),
    * @example DateTime.fromObject({ hour: 10, minute: 26, second: 6 }, { zone: 'local' })
@@ -766,7 +768,9 @@ export default class DateTime {
     const normalized = normalizeObject(obj, normalizeUnitWithLocalWeeks);
     const { minDaysInFirstWeek, startOfWeek } = usesLocalWeekValues(normalized, loc);
 
-    const tsNow = Settings.now(),
+    const tsNow = isUndefined(opts.refrenceDate)
+      ? friendlyDateTime(opts.refrenceDate).toUnixInteger()
+      : opts.refrenceDate,
       offsetProvis = !isUndefined(opts.specificOffset)
         ? opts.specificOffset
         : zoneToUse.offset(tsNow),
--- a/test/datetime/create.test.js
+++ b/test/datetime/create.test.js
@@ -883,6 +883,12 @@ test("DateTime.fromObject takes a undefined to mean {}", () => {
   expect(res.year).toBe(new Date().getFullYear());
 });
 
+test("DateTime.fromObject respects `refrenceDate`", () => {
+  const res = DateTime.fromObject(undefined, { refrenceDate: { day: 10 } });
+  expect(res.year).toBe(new Date().getFullYear());
+  expect(res.day).toBe(10);
+});
+
 test("private language subtags don't break unicode subtags", () => {
   const res = DateTime.fromObject(
     {},


src/datetime.js
```diff
@@ -745,8 +745,10 @@ export default class DateTime {
    * @param {string} [opts.locale='system\'s locale'] - a locale to set on the resulting DateTime instance
    * @param {string} opts.outputCalendar - the output calendar to set on the resulting DateTime instance
    * @param {string} opts.numberingSystem - the numbering system to set on the resulting DateTime instance
+   * @param {DateTime|Date|Object} opts.refrenceDate - the reference date to taken for missing parts
```
dasa, line 748:
```suggestion
   * @param {DateTime|Date|Object} opts.refrenceDate - the reference date to take for missing parts
```

angelaki: Whoops ... Totally with you

angelaki:
So what are we waiting for? :)

dasa:
I'm not sure about the logic, but there is a spelling typo I mentioned at: https://github.com/moment/luxon/pull/1632#issuecomment-2118162985

src/datetime.js
```diff
@@ -766,7 +768,9 @@ export default class DateTime {
     const normalized = normalizeObject(obj, normalizeUnitWithLocalWeeks);
     const { minDaysInFirstWeek, startOfWeek } = usesLocalWeekValues(normalized, loc);
 
-    const tsNow = Settings.now(),
+    const tsNow = isUndefined(opts.refrenceDate)
+      ? friendlyDateTime(opts.refrenceDate).toUnixInteger()
+      : opts.refrenceDate,
```
dasa, line 773: I'm not a member of the Luxon team, but here's my thoughts:

- maybe `tsNow` would be better named `tsRef`
- I don't think `opts.refrenceDate` should be used if it's `undefined`

angelaki:
Fixed your concerns. Thank you pretty much. Time for a merge now? @Luxon-Team?

src/datetime.js
```diff
@@ -745,8 +745,10 @@ export default class DateTime {
    * @param {string} [opts.locale='system\'s locale'] - a locale to set on the resulting DateTime instance
    * @param {string} opts.outputCalendar - the output calendar to set on the resulting DateTime instance
    * @param {string} opts.numberingSystem - the numbering system to set on the resulting DateTime instance
+   * @param {DateTime|Date|Object} opts.referenceDate - the reference date to take for missing parts
    * @example DateTime.fromObject({ year: 1982, month: 5, day: 25}).toISODate() //=> '1982-05-25'
    * @example DateTime.fromObject({ year: 1982 }).toISODate() //=> '1982-01-01'
+   * @example DateTime.fromObject({ year: 1982 }, { referenceDate: { day: 10 } }).toISODate() //=> '1982-01-10'
```
diesieben07, line 751: This example is actually wrong. The referenceDate is only used for units above the highest unit in the input. If `{ month: 6 }` is given, _only_ the year is taken from the reference date.

test/datetime/create.test.js
```diff
@@ -883,6 +883,12 @@ test("DateTime.fromObject takes a undefined to mean {}", () => {
   expect(res.year).toBe(new Date().getFullYear());
 });
 
+test("DateTime.fromObject respects `referenceDate`", () => {
+  const res = DateTime.fromObject(undefined, { referenceDate: { day: 10 } });
+  expect(res.year).toBe(new Date().getFullYear());
+  expect(res.day).toBe(10);
+});
+
```
diesieben07, line 891: I think it would be good to also add tests for:
- Calling `fromObject` with actual data and check that the data is "merged in".
- Checking that `referenceDate` is only used for units higher than the highest unit in the source object (i.e. `fromObject({ day: 10 })` reads _only_ `year` and `month` from the reference date.
- Some tests for methods that indirectly call `fromObject` (like `fromISO`) with `referenceDate`.

src/datetime.js
```diff
@@ -745,8 +745,10 @@ export default class DateTime {
    * @param {string} [opts.locale='system\'s locale'] - a locale to set on the resulting DateTime instance
    * @param {string} opts.outputCalendar - the output calendar to set on the resulting DateTime instance
    * @param {string} opts.numberingSystem - the numbering system to set on the resulting DateTime instance
+   * @param {DateTime|Date|Object} opts.referenceDate - the reference date to take for missing parts
```
diesieben07, line 748: Please add the documentation for this parameter on all factory methods that ultimately call `fromObject`, like `fromISO`.

diesieben07 (changes requested):

src/datetime.js
angelaki: Oh sorry, I just expected the _old_ `tsNow` to add all the missing parts (not just the units above). What's your opinion here? Change the logic on a passed refDate, so **all** missing parts are replaced? And if, always (could be a breaking change?) or just with passed refDate?

date-fns even uses proximity to guess the right input: https://date-fns.org/v3.6.0/docs/parse

src/datetime.js
angelaki: Will do 🫡 as soon as we have a solution for the refDate behavior

angelaki: Hmmm ... just wanted to start add some. But since your pretty much more into luxon ... could you please support me here? Would need to check the whole code, guess you know all the methods?

src/datetime.js
diesieben07: If that is added, it definitely needs to be opt in. `DateTime.fromObject({ year: 2013 })` shouldn't suddenly have the current time.
What's your use-case for this?

src/datetime.js
diesieben07: This should be all of the parsers:
image

src/datetime.js
angelaki: Tbh, I have no use-case for it. Just thinking, in an Englisch format, if only this first digits (month) is set, maybe someone wants to have a different day than the first (and not only a different year than todays?). But again, no hard feeling here! Can just stay as it is!

diesieben07: Let's keep it as it is for now then!

diesieben07:
I'll do another review pass this evening!

angelaki:
@diesieben07 did _this evening_ already happen? ;) Sorry for bumping, no hurries! Just kidding :)

src/datetime.js
diesieben07: The example here still contradicts the actual behavior and implementation.

diesieben07 (changes requested):
I apologize for the delay on getting back to you here.
Unfortunately it seems you haven't really addressed the problems I highlighted in my previous review regarding tests and the wrong example.

diesieben07:
> Just wanted to create a PR for a twoDigitCutoffYear parameter (for Settings being just used as fallback) but used my master branch for this fix. So it's kinda unhandy to create a new PR. My fault ...

You can still create a new branch in your repo that tracks Luxon's master branch, they don't have to be named the same.

PR #1642, opened 2024-06-27
Optimize Performance with Caching for Resolved Options and Parsed Locale Strings

### LINKED ISSUE #1428: [perf] isEnglish and supportsFastNumbers are slow - `new Intl.DateTimeFormat`
I'm profiling a calendar page which shows a bunch of events. I collect a profiling file and open it in https://www.speedscope.app, look at sandwich view, and see that `isEnglish` is the 3th slowest function (by self time) (and supportsFastNumbers is 5th):
image

I take a look at `isEnglish` :
https://github.com/moment/luxon/blob/07023283e74321c2080de9bb73070a3ffd2a7acc/src/impl/locale.js#L479-L485

As I'm working in Norwegian locale, I assume this returns false, but most importantly - creates `new Intl.DateTimeFormat` on every call.

I have not tried this yet, but I believe we can win some time if we cache `Intl.DateTimeFormat` (or avoid doing this check).

Same happens in `supportsFastNumbers`.

- OS: Win 11
- Browser: Chrome 112.0.5615.138
- Luxon: 3.3.0

nodify-at:
#### Changes
* **Caching:** Implemented caching to avoid redundant computations of resolved options and locale parsing.
* **Efficient Checks:** Introduced a final isEnglish variable to efficiently determine if the locale supports English.
* **Initialization**: supportsFastNumbers is now immediately initialized to prevent unnecessary calls.

#### Performance Impact
Tests indicate up to a 10x speed improvement with these enhancements.

#### Issue Fixed
Fixes #1428

### DIFF
--- a/src/impl/locale.js
+++ b/src/impl/locale.js
@@ -62,7 +62,7 @@ function systemLocale() {
 }
 
 let resolvedOptionsCache = {};
-function resolvedOptions(locString) {
+function getCachedIntResolvedOptions(locString) {
   if (!resolvedOptionsCache[locString]) {
     resolvedOptionsCache[locString] = new Intl.DateTimeFormat(locString).resolvedOptions();
   }
@@ -81,18 +81,7 @@ function getCachedWeekInfo(locString) {
   return data;
 }
 
-const localeStringCache = {};
-function getParsedCachedLocaleString(localeStr) {
-  if (localeStringCache[localeStr]) {
-    return localeStringCache[localeStr];
-  }
-
-  const result = _parseLocaleString(localeStr);
-  localeStringCache[localeStr] = result;
-  return result;
-}
-
-function _parseLocaleString(localeStr) {
+function parseLocaleString(localeStr) {
   // I really want to avoid writing a BCP 47 parser
   // see, e.g. https://github.com/wooorm/bcp-47
   // Instead, we'll do this:
@@ -129,13 +118,7 @@ function _parseLocaleString(localeStr) {
   }
 }
 
-const intlConfigCache = {};
 function intlConfigString(localeStr, numberingSystem, outputCalendar) {
-  const key = `${localeStr}${numberingSystem}${outputCalendar}`;
-  if (intlConfigCache[key]) {
-    return intlConfigCache[key];
-  }
-
   if (outputCalendar || numberingSystem) {
     if (!localeStr.includes("-u-")) {
       localeStr += "-u";
@@ -148,9 +131,10 @@ function intlConfigString(localeStr, numberingSystem, outputCalendar) {
     if (numberingSystem) {
       localeStr += `-nu-${numberingSystem}`;
     }
+    return localeStr;
+  } else {
+    return localeStr;
   }
-  intlConfigCache[key] = localeStr;
-  return localeStr;
 }
 
 function mapMonths(f) {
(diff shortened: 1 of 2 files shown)


nodify-at:
Is there any update on this or can someone give me a feedback if I've to improve this PR? :)

flodlc:
Once this one is merged, could we expect a new release ? There has been many perf improvements that would benefits to the community.

icambron:
I'll take a look at this one soon

nodify-at:
Thanks @icambron :)

Seledrex:
Nice work! I was just investigating this problem myself and stumbled upon this issue. Definitely looking forward to this performance improvement! 🚀

src/impl/locale.js
```diff
@@ -374,14 +394,14 @@ export default class Locale {
     this.eraCache = {};
 
     this.specifiedLocale = specifiedLocale;
-    this.fastNumbersCached = null;
+    this.fastNumbersCached = supportsFastNumbers(this);
```
icambron, line 397: I guess I don't quite know what moving this here is fixing

src/impl/locale.js
```diff
@@ -374,14 +394,14 @@ export default class Locale {
...
+    this.isEnglishCached =
```
icambron, line 398: Similarly, I agree we should cache this, but why not cache it inside `isEnglish`. That way if `isEnglish` is never called, we don't pay the performance penalty

src/impl/locale.js
```diff
@@ -123,10 +148,9 @@ function intlConfigString(localeStr, numberingSystem, outputCalendar) {
     if (numberingSystem) {
       localeStr += `-nu-${numberingSystem}`;
     }
-    return localeStr;
-  } else {
-    return localeStr;
   }
+  intlConfigCache[key] = localeStr;
```
icambron, line 152: I'm having a hard time understanding how this would give such a big speedup. Can you use the benchmarks in the repo to demonstrate this?

src/impl/locale.js
```diff
@@ -73,7 +81,18 @@ function getCachedWeekInfo(locString) {
   return data;
 }
 
-function parseLocaleString(localeStr) {
+const localeStringCache = {};
+function getParsedCachedLocaleString(localeStr) {
```
icambron, line 85: I can believe that this parsing is slow, but I'd like to see the benchmarks on this too

icambron (changes requested):
Sorry for taking so long to review this. I left some comments

nodify-at:
@icambron Thanks for your feedback. I took a thorough approach initially to improve the calls, but I've now simplified it to focus on the isEnglish and fastNumbers calls only. I've also cleaned up all other improvements/changes. Could you please review it again?

nodify-at:
@icambron 
I hope I did this correctly. I also added two new benchmark tests to compare the results.

Here are the initial results from my local environment:

DateTime#format in German: 301,824 ops/sec ±0.41% (95 runs sampled)
DateTime#format in German with no-cache: 11,999 ops/sec ±3.03% (88 runs sampled)

nodify-at:
@icambron did you get my last changes ?

flodlc:
I'm starting to lose hope that this PR will ever get merged 😞

Seledrex:
> I'm starting to lose hope that this PR will ever get merged 😞 

Just give it some more time!

icambron (approved):

icambron:
Looks good, merged. Thanks for the contribution. I will release this soon.

nodify-at:
Awesome :) Thanks @icambron

src/impl/locale.js
```diff
@@ -167,7 +175,7 @@ function supportsFastNumbers(loc) {
       loc.numberingSystem === "latn" ||
       !loc.locale ||
       loc.locale.startsWith("en") ||
-      new Intl.DateTimeFormat(loc.intl).resolvedOptions().numberingSystem === "latn"
+      getCachedIntResolvedOptions(loc.locale).numberingSystem === "latn"
```
mohd-akram, line 178: Should this be `loc.intl`?

nodify-at: It looks a bit strange, because in my last commit I reverted this change. getCachedIntResolvedOptions function is used only in supportsFastNumbers and isEnglish .

See the code in master: https://github.com/moment/luxon/blob/master/src/impl/locale.js

nodify-at: Ops, I see your pointed code better now, sorry and you're right it should be intl.

nodify-at: https://github.com/moment/luxon/pull/1654

PR #1643, opened 2024-06-28
fix: Allow default initialization of DateTime/Duration/Interval

mbStavola:
This PR default initializes the config parameter of the constructor for `DateTime`, `Duration`, and `Interval` in order to fix the following:

```
Cannot read properties of undefined (reading 'zone')
```

This can be encountered when using a third party serialization/deserialization library such as class-transformer as temporary objects may be constructed with `new` while waiting for any transformation logic to run. This is actually a very common issue for NestJS users.

Understandably, these constructors are *meant* to be private, but the unfortunate fact is that nothing can prevent `new` from being called on these types. Moreover, if you look at the constructor definitions it seems that the intent was to provide sane defaults *already*. With that in mind, it may make sense just to provide a default to sidestep the issue entirely and let the existing code handle things.

Ultimately, this is a minor adjustment which fixes some very real pain points.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -514,6 +514,10 @@ export default class DateTime {
    * @access private
    */
   constructor(config) {
+    if (!config) {
+      return DateTime.invalid("no config supplied");
+    }
+
     const zone = config.zone || Settings.defaultZone;
 
     let invalid =
--- a/src/duration.js
+++ b/src/duration.js
@@ -218,6 +218,10 @@ export default class Duration {
    * @private
    */
   constructor(config) {
+    if (!config) {
+      return Duration.invalid("no config supplied");
+    }
+
     const accurate = config.conversionAccuracy === "longterm" || false;
     let matrix = accurate ? accurateMatrix : casualMatrix;
 
--- a/src/impl/locale.js
+++ b/src/impl/locale.js
@@ -360,7 +360,9 @@ export default class Locale {
   }
 
   constructor(locale, numbering, outputCalendar, weekSettings, specifiedLocale) {
-    const [parsedLocale, parsedNumberingSystem, parsedOutputCalendar] = parseLocaleString(locale);
+    const [parsedLocale, parsedNumberingSystem, parsedOutputCalendar] = parseLocaleString(
+      locale || systemLocale()
+    );
 
     this.locale = parsedLocale;
     this.numberingSystem = numbering || parsedNumberingSystem || null;
--- a/src/interval.js
+++ b/src/interval.js
@@ -41,6 +41,10 @@ export default class Interval {
    * @private
    */
   constructor(config) {
+    if (!config) {
+      return Interval.invalid("no config supplied");
+    }
+
     /**
      * @access private
      */
--- a/test/datetime/create.test.js
+++ b/test/datetime/create.test.js
@@ -10,6 +10,15 @@ const withDefaultLocale = Helpers.withDefaultLocale,
   withDefaultZone = Helpers.withDefaultZone;
 
 //------
+// new
+//------
+test("Default constructor works and returns an invalid instance", () => {
+  const datetime = new DateTime();
+  expect(datetime.isValid).toBe(false);
+  expect(datetime.invalidReason).toBe("no config supplied");
+});
+
+//------
 // .now()
 //------
 test("DateTime.now has today's date", () => {
--- a/test/duration/create.test.js
+++ b/test/duration/create.test.js
@@ -3,6 +3,15 @@
 import { Duration } from "../../src/luxon";
 
 //------
+// new
+//------
+test("Default constructor works and returns an invalid instance", () => {
+  const duration = new Duration();
+  expect(duration.isValid).toBe(false);
+  expect(duration.invalidReason).toBe("no config supplied");
+});
+
+//------
 // .fromObject()
 //-------
 test("Duration.fromObject sets all the values", () => {
--- a/test/interval/create.test.js
+++ b/test/interval/create.test.js
@@ -5,6 +5,15 @@ import Helpers from "../helpers";
 const withThrowOnInvalid = Helpers.setUnset("throwOnInvalid");
 
 //------
+// new
+//------
+test("Default constructor works and returns an invalid instance", () => {
+  const interval = new Interval();
+  expect(interval.isValid).toBe(false);
+  expect(interval.invalidReason).toBe("no config supplied");
+});
+
+//------
 // .fromObject()
 //-------
 test("Interval.fromDateTimes creates an interval from datetimes", () => {


mbStavola:
I am also aware of [this previous attempt](https://github.com/moment/luxon/pull/1466), but I did not want to revive a year old dead PR. Additionally, that PR only covers `DateTime`.

mbStavola:
@icambron just saw the force-push notification here, was wondering if you had some time to review this?

icambron:
Hmm, do the objects you construct this way actually work? I'm sort of guessing you'll get errors using them because e.g. `values` is not defined...

mbStavola:
@icambron Would it be better to instead have the constructor return invalid instances instead of a config is not supplied? That way construction doesn't throw an error, but the object returned is properly marked as invalid? For example, something like:

```javascript
export default class DateTime {
  /**
   * @access private
   */
  constructor(config) {
    if (!config) {
      return DateTime.invalid('no config supplied');
    }
```

Something like this would still require that a config is supplied in order to create valid instances, but save some headaches for those using libraries which by default `new` these objects.

mbStavola:
For those interested, here are some of the simpler test cases I ran to confirm the changes:

```javascript
/* global test expect */

// Requires an npm i -D class-transformer
import { plainToInstance, instanceToPlain, Transform } from "class-transformer";

import { DateTime } from "../src/luxon";

export const IntoDateTimeTransformer = (
  { format, acceptNulls, ...opts } = {},
  options = {},
) => {
  return Transform(
    ({ value }) => {
      if (DateTime.isDateTime(value)) {
        return value;
      }

      return DateTime.fromISO(value, { setZone: true, ...opts });
    },
    { ...options, toClassOnly: true },
  );
};

export const FromDateTimeTransformer = (
  { format, acceptNulls, ...opts } = {},
  options = {},
) => {
  return Transform(
    ({ value }) => {
      if (!DateTime.isDateTime(value)) {
        throw new Error('Value is not a DateTime');
      }

      return value.toISO(opts);
    },
    { ...options, toPlainOnly: true },
  );
};

// This is what Typescript will codegen as a decorator helper function
// Stealing it to make our lives easier in testing
const __decorate = (this && this.__decorate) || function (decorators, target, key, desc) {
  var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
  if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
  else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
  return c > 3 && r && Object.defineProperty(target, key, r), r;
};

class Person {
  name;
  birthday;

  constructor(name, birthday) {
    this.name = name;
    this.birthday = birthday;
  }
}

__decorate([
  IntoDateTimeTransformer(),
  FromDateTimeTransformer()
], Person.prototype, "birthday", undefined);

test("Instance to plain works", () => {
  const person = new Person('Bob', DateTime.fromISO('1990-10-12', { zone: 'UTC' }));
  const serialized = instanceToPlain(person);

  expect(serialized.name).toBe('Bob');
  expect(serialized.birthday).toBe('1990-10-12T00:00:00.000Z');
});

test("Plain to instance works (plain string)", () => {
  const person = { name: 'Alice', birthday: '1992-05-15T00:00:00.000Z' };
  const deserialized = plainToInstance(Person, person);

  expect(deserialized.name).toBe('Alice');
  expect(deserialized.birthday).toStrictEqual(DateTime.fromISO('1992-05-15', { zone: 'UTC' }));
});

test("Plain to instance works (datetime object)", () => {
  const person = { name: 'Eve', birthday: DateTime.fromISO('1991-08-03', { zone: 'UTC' }) };
  const deserialized = plainToInstance(Person, person);

  expect(deserialized.name).toBe('Eve');
  expect(deserialized.birthday).toStrictEqual(DateTime.fromISO('1991-08-03', { zone: 'UTC' }));
});
```

On the primary branch, that last test will fail due to the aforementioned issue. I did not commit this because I don't think its the responsibility of Luxon to test for class-transformer compatibility _specifically_, however I did include tests in this MR to ensure the default constructor works.

mbStavola:
@icambron Small nudge, I know you're probably very busy but any chance you can take another look?

mbStavola:
@diesieben07 @icambron One last bump, any chance for a review?

diesieben07:
Sorry, but I am not willing to open up the constructor as public API (which this effectively does).
The only API to construct Luxon objects is via the factory functions.
My plan for the next major version is to make the constructors _truly_ private (i.e. they will throw when called from outside Luxon).

You should instruct your serialization/deserialization library for how to interact with Luxon objects, that is, serialize via something like toISO and deserialize via the corresponding factory function.
Can you elaborate on why this is not possible?

mbStavola:
@diesieben07 I can elaborate a bit-- the motivating example in `class-transfomers` is pretty specific so I don't blame you for thinking that this is a strange ask. To be completely honest, I think most people who encounter this sort of issue end up pretty confused as to what is happening and figuring out why can be a bit of a journey.

Anyway, if you look at [the test](https://github.com/moment/luxon/pull/1643#issuecomment-2394729007) I wrote up, you can actually see where I "instruct" the serializer and deserializer to use `toISO` and `fromISO` respectively, as well as having an early-out path if we're _already_ a `DateTime` (this will be important shortly). This _can_ work, but it won't always; in certain scenarios `class-transformer` will call `new` on the target type that it expects to serde against, pretty much as a placeholder or default.

The last test is an example of one of those scenarios: deserializing a type that _already has a `DateTime` in it_, something that should work given that early-out from before. Often this is encountered when shuttling already deserialzed models around after parsing a request (e.g. caching, mqs) and its especially prominent when doing testing. Of course I am only using `class-transformer` as an example here because it's likely the most popular library which runs into this (as its used by NestJS), but as the original issue points out other mapper-like libraries are susceptible.

Regardless, I'm very sympathetic to having a private constructor and encouraging the usage of factory functions to engage with the API as it's much cleaner and easier to understand. However, this does mean that the door is effectively closing on a solution here. Perhaps its wishful thinking, but I'm sure a lot of affected users would be happy to even have something like a `Settings.throwOnInvalid` equivalent to bypass default constructor protections.

Anyway, thanks for you reply and thanks for your effort working on Luxon!

diesieben07:
> in certain scenarios class-transformer will call new on the target type that it expects to serde against, pretty much as a placeholder or default.

This sounds like a bug in class-transfomer to me, unless they explicitly state that you can only use it with types that can just "randomly be new-ed". I haven't used class-transformer, is this behavior documented anywhere? Can you elaborate on _when_ it occurs?
I really would prefer finding a solution to this that does not involve allowing `new DateTime` as an API to construct a DateTime instance that must behave correctly.

diesieben07:
Okay I have dug into this more. This is, in my opinion, 100% a bug in class transformer.
Here is what's happening when you do this:
```ts
plainToInstance(User, {
  id: 1,
  name: "John Doe",
  birthday: DateTime.fromISO("2000-01-01")
});
```

1. It constructs an instance of the target class (`new User()`).
2. It loops over the keys of the source object (id, name, birthday) and transforms each of them in order.
3. When it gets to "birthday", it finds that the source value is an object (a `DateTime` instance) and not one of the special types class transformer handles (say `Date` or `Array`).
4. This code path for "source value is an object" now does the same class-transformer process, by new-ing the target class and then copying over properties. This happens even if you have a `@Transform` on the property with `toClassOnly: true`. This is where the bug is, in my opinion.
4. After it has created a new `DateTime` and then copied over all the properties it can find it will pass _that_ value to the `@Transform`.

This breaks _entirely_ if `DateTime` were to use private properties (`#` prefixed), because class transformer can't see them and its copying over of properties would do nothing (since there are no public properties) and then the instance that `@Transform` receives would be completely broken.

This is something that class transformer needs to fix, not Luxon. As it stands, class transformer can only be used with classes that are specifically designed for it and Luxon's classes are not.
And I am not willing to make "create a new instance of `DateTime` with no arguments and then copy over random properties" a supported API. It is already on my list to make the internal properties of `DateTime` (such as `_zone` or `c`) [actually private](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties), which would break this again, even with your patch.

I have created a demo project to show this problematic behavior:
https://github.com/diesieben07/class-transformer-test

1. `npx tsx index.ts` shows that it "works" with a normal property, but traces where the problematic constructor call happens.
2. `npx tsx index-private-property.ts` shows that it completely breaks when the class uses private properties.

Feel free to do with this what you will, but I am not willing to make Luxon support this broken way of constructing its classes.

PR #1660, opened 2024-10-03
Use techie offset for basic ISO format

snowwwz:
# Description of the Bug

with `format: 'basic'` option, `toISO` and `toISOTime` is supposed to return techie offset (ex. `+0900`)
 
https://github.com/moment/luxon/blob/3e9983cd0680fdf7836fcee638d34e3edc682380/src/datetime.js#L1825

but, it returns short offset (ex. `+09:00`) with `:` instead.

```js
% node
Welcome to Node.js v22.3.0.
Type ".help" for more information.
> const { DateTime } = await import("luxon");
undefined
> DateTime.now().toISO({ format: 'basic' });
'20241003T231708.609+09:00'
```

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -266,12 +266,16 @@ function toISOTime(
     } else if (o.o < 0) {
       c += "-";
       c += padStart(Math.trunc(-o.o / 60));
-      c += ":";
+      if (extended) {
+        c += ":";
+      }
       c += padStart(Math.trunc(-o.o % 60));
     } else {
       c += "+";
       c += padStart(Math.trunc(o.o / 60));
-      c += ":";
+      if (extended) {
+        c += ":";
+      }
       c += padStart(Math.trunc(o.o % 60));
     }
   }
--- a/test/datetime/format.test.js
+++ b/test/datetime/format.test.js
@@ -247,6 +247,7 @@ test("DateTime#toISOTime() can omit the offset", () => {
 
 test("DateTime#toISOTime() can output the basic format", () => {
   expect(dt.toISOTime({ format: "basic" })).toBe("092354.123Z");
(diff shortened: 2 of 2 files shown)


mceachen:
Why would you make this change? It breaks prior default behavior and will almost certainly cause downstream issues.

Technically, ISO8601 states that the colon is optional, the default ISO format includes the colon. See https://en.wikipedia.org/wiki/ISO_8601

dobon:
@mceachen This PR does not alter the default behavior because the default format is 'extended', not 'basic'. This PR corrects a regression that occurred between these two commits:
**before:** https://github.com/moment/luxon/commit/8b65d23da929ec52b753e98543f8f2d02ec09e65#diff-5146f40a8c8111b14e5ab3407eff80c7a6d2f3c92a4bf4ccd4647bff32d47e38L219
(NB: if using 'basic' format, renders zone offset using `ZZZ`, which omits the colon, and otherwise uses `ZZ`, which includes the colon.)
**after:** https://github.com/moment/luxon/pull/1108/commits/fac04675f6b88472d511c55770de93124cc46ca2
(NB: this commit creates the toIsoTime method and obliterates the 'basic' format behavior by hard-coding the colons in zone offset rendering logic.)

Reading the notes on that 'after' PR pointed me to https://github.com/moment/luxon/pull/1097/files which shows the logic that was omitted by mistake in the regression commit above.

@diesieben07 We use similar logic to this commit throughout the `toIsoTime` and `toIsoDate` methods, and this corrects a genuine bug in our ISO basic format rendering, so I think this PR should be merged. My only very slight misgiving is that it checks `format` twice which could be trivially reduced to once, and that minor optimization might be worthwhile given the performance priority of this method. What do you think?

PR #748, opened 2020-07-29
To iso truncate time

nbcarey:
This addresses Issue #651 (_Feature request - function to quickly create ISO8601 stamp without milliseconds_).

It adds a `precision` option, the domain of which is `hours`, `minutes`, `seconds`, `milliseconds`, defaulting to `milliseconds`. Time components of higher precision than that specified are simply omitted: no attempt is made at rounding.

If `precision` is specified in addition to `suppressSeconds` or `suppressMilliseconds`, the `suppress*` option wins:
```javascript
DateTime.utc({2020,7,25,12,34,0,0}).toISO({ precision: 'millisecond',  suppressSeconds: true })
```
returns `2020-07-25T12:34Z`.

Millisecond precision was requested, with the proviso that seconds and milliseconds won't be rendered if they are both zero.

**Note:** It might be a nice TO DO to add an `auto` precision that would set the precision level to omit trailing components whose value is zero, so
```javascript
Datetime.utc({2020, 7, 25, 12, 34, 0, 0}).toISO({ precision: "auto" })
```
 would return `12:34Z`.

Issue #651, opened 2020-01-21
Feature request - function to quickly create ISO8601 stamp without milliseconds


cyberwombat:
Unless I am missing something this is how one gets the DateTime without milliseconds:
```
DateTime.utc()
        .startOf('second')
        .toISO({ suppressMilliseconds: true }),
```
Which is rather verbose. So many systems use the ISO8601 format without milliseconds (ex: 2020-01-21T23:03:23Z) that it would be nice to have a faster way to crank these out? maybe a `trimMilliseconds` option or something.

And thanks for the lib - I've tried them all and love this one by far the most.

icambron:
I'd accept a PR for this

cyberwombat:
Coo.. What do you think of creating a dedicated function `toISODateTime()`. ? else I think it would be an option like `trimMilliseconds` though that could create a bit of confusion between that and `suppressMuliseconds'

icambron:
I like the option better. If it's confusing, perhaps `suppressMilliseconds` needs a better name

ahmetzambak:
I'm missing that feature too

EzeGmnz:
The following PR solves this by adding a new parameter https://github.com/moment/luxon/pull/1412

TheArKaID:
@EzeGmnz  Is this not accepted, or am I missing a docs that mentions it?

cyberwombat:
@TheArKaID  its in there. Perhaps not documented - is an option for `toISO` and `toISOTime` on DateTime as well as in Duration methods. See code: https://github.com/search?q=repo%3Amoment%2Fluxon+suppressMilliseconds&type=code

TheArKaID:
> @TheArKaID its in there. Perhaps not documented - is an option for `toISO` and `toISOTime` on DateTime as well as in Duration methods. See code: https://github.com/search?q=repo%3Amoment%2Fluxon+suppressMilliseconds&type=code

Thanks for mentioning it. But, it can only be suppressed if it is zero, isn't?
![image](https://github.com/moment/luxon/assets/35050722/8c83d44b-4d01-47a2-abec-cabe301a39d7)

cyberwombat:
I can’t remember for sure, but you can certainly round down or upOn Jun 2, 2024, at 9:52 AM, Arifia Kasastra R ***@***.***> wrote:﻿

[...]

PR #1670, opened 2024-12-05
Create ISO 8601 Date/Time Values w/o Milliseconds (re-implemented, ready for review)

dobon:
This PR duplicates and updates PR: https://github.com/moment/luxon/pull/748 and issue: https://github.com/moment/luxon/issues/651

I've incorporated the feedback from that review and PR, minus renaming "supress[milli]second" which should be in a different PR if the change is made.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -214,21 +214,14 @@ function toTechFormat(dt, format, allowZ = true) {
     : null;
 }
 
-function toISODate(o, extended) {
-  const longFormat = o.c.year > 9999 || o.c.year < 0;
-  let c = "";
-  if (longFormat && o.c.year >= 0) c += "+";
-  c += padStart(o.c.year, longFormat ? 6 : 4);
+function toISODate(o, extended, precisionUnitIndex = 2) {
+  const extendedGlyph = extended ? "-" : "",
+    yearExceedsFourDigits = o.c.year > 9999,
+    yearPadding = o.c.year < 0 || yearExceedsFourDigits ? 6 : 4;
 
-  if (extended) {
-    c += "-";
-    c += padStart(o.c.month);
-    c += "-";
-    c += padStart(o.c.day);
-  } else {
-    c += padStart(o.c.month);
-    c += padStart(o.c.day);
-  }
+  let c = (yearExceedsFourDigits ? "+" : "") + padStart(o.c.year, yearPadding);
+  if (precisionUnitIndex >= 1) c += extendedGlyph + padStart(o.c.month);
+  if (precisionUnitIndex >= 2) c += extendedGlyph + padStart(o.c.day);
   return c;
 }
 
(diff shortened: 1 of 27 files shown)


src/datetime.js
```diff
@@ -1819,10 +1823,12 @@ export default class DateTime {
    * @param {boolean} [opts.includeOffset=true] - include the offset, such as 'Z' or '-04:00'
    * @param {boolean} [opts.extendedZone=false] - add the time zone format extension
    * @param {string} [opts.format='extended'] - choose between the basic and extended format
+   * @param {string} [opts.precision='milliseconds'] - specify desired time precision: 'hours', 'minutes', 'seconds' or 'milliseconds'
```
diesieben07, line 1826: This should mention
- that it will _truncate_ the output
- How it interacts with `suppressSeconds` and `suppressMilliseconds`

diesieben07 (changes requested):
Should we also support this for calendar units? 
`2024-05` is a valid ISO date, referring to the month of may 2025.

dobon:
> Should we also support this for calendar units? `2024-05` is a valid ISO date, referring to the month of may 2025.

There was discussion about this in the prior PR, but no conclusion. I've added another commit to enable precision from Year to Millisecond, and it seems good to me. Do the test cases show the function behaving the way you had in mind @diesieben07 ?

Also, I still need to update the method comments to specify truncation behavior and suppressSeconds / suppressMilliseconds behavior (your other comment). I'll include a no-code patch for that in the coming days.

src/datetime.js
```diff
@@ -1844,33 +1831,42 @@ export default class DateTime {
...
-  toISODate({ format = "extended" } = {}) {
-    if (!this.isValid) {
-      return null;
-    }
-
-    return toISODate(this, format === "extended");
+  toISODate({ format = "extended", precision = "day" } = {}) {
+    if (!this.isValid) return null;
+    const precisionUnitIndex = orderedUnits.indexOf(normalizeUnit(precision));
+    return toISODate(this, format === "extended", precisionUnitIndex);
```
dobon, line 1869: add `if (precisionUnitIndex < 0) throw new InvalidUnitError(precision)` here... except somehow the test is already passing, so investigate why that test is either not working, or else why the system is somehow throwing the correct error anyway.

dobon: the error throwing was being handled by the `normalizeUnit` method, so I removed the superfluous error throwing logic.

src/datetime.js
```diff
@@ -238,26 +245,57 @@ function toISOTime(
...
-  if (o.c.millisecond !== 0 || o.c.second !== 0 || !suppressSeconds) {
-    c += padStart(o.c.second);
-
-    if (o.c.millisecond !== 0 || !suppressMilliseconds) {
-      c += ".";
-      c += padStart(o.c.millisecond, 3);
-    }
+  let showSeconds,
+    c = "";
+  switch (precisionUnitIndex) {
```
diesieben07, line 253: I'm not a fan of the code duplication as well as the magic index numbers here.

diesieben07 (changes requested):

src/datetime.js
dobon: Thanks for the feedback.

Both the loop unraveling(code duplication) and magic index numbers were implemented like that for performance (ie: to minimize comparisons) because toISO can be a hot path.

The loop unravelling was to avoid unnecessary compares on smaller units (eg: if precision is 'days', then don't bother checking if precision is greater than years and months), and the magic index numbers were to facilitate numeric compares instead of string.

That being said, if performance isn't paramount then I agree that the repetitive code reads not-great and I'm happy to re-implement any way you please. As such, I've pushed a commit that passes precision as a string parameter (eliminating the magic numbers), and eliminates the loop unraveling in exchange for a couple more compares. Let me know what you think.

diesieben07: I agree that `toISO` should be fast. But we shouldn't make claims about performance without a benchmark :D

diesieben07:
Thank you!

Issue #1644, opened 2024-07-09
Unclear Documentation for toSeconds() Function


reid-wheeler:
**Is your feature request related to a problem? Please describe.**
It is unclear in the documentation that the `toSeconds()` function returns a floating point number.

**Describe the solution you'd like**
Update to documentation to read: "Returns the epoch seconds (as a floating point number) of this DateTime."

**Describe alternatives you've considered**
N/A

**Additional context**
Most datetime libraries return a whole number to represent seconds from Epoch. It would clear up some confusion if the documentation for the `toSeconds()` function mentioned that it returns a floating point number. `toUnixInteger()` specifies that it is a whole number, and I believe `toSeconds()` should make a similar distinction.

1dylan1:
Relevant previous discussion on same topic here for this interested: https://github.com/moment/luxon/issues/565

I've submitted a Pull request here that updates the docstring:  https://github.com/moment/luxon/pull/1681

reid-wheeler:
> Relevant previous discussion on same topic here for this interested: [#565](https://github.com/moment/luxon/issues/565)
> 
> I've submitted a pull request here that updates the docstring to avoid any further confusion: [#1681](https://github.com/moment/luxon/pull/1681)

Thank you!

PR #1681, opened 2025-01-31
Include documentation that toSeconds() returns a floating point number

1dylan1:
Relevant discussion here: https://github.com/moment/luxon/issues/565

Newest discussion was brought up again here, indicating users are still having issues with toSeconds() returning the milliseconds as the floating point part of the number: https://github.com/moment/luxon/issues/1644

I've updated the doc string to match the toUnixInteger's format of indicating it returns a floating point instead of a whole number.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -2030,7 +2030,7 @@ export default class DateTime {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


diesieben07:
Thank you!

Issue #1129, opened 2022-01-28
More control over DateTime#toRelative rounding behavior


corymharper:
**Is your feature request related to a problem? Please describe.**
For some units of time, rounding down is very unintuitive. It's probably fine if 29.99 minutes rounds down to 29, but it's fairly inaccurate to describe 1.99 years away as "1 year away." Padding in these situations that could be minutes away or years away can lead to a lot of extra code to figure out how much padding there should be in this situation.

**Describe the solution you'd like**
I think one possible solution would be to change the `round` option you can pass to `toRelative` from a on/off style boolean into a setting probably based on a string. Something like `round: 'down' | 'up' | 'closest' | 'off'` where, if we want to keep the current default, it defaults to `down`. `off` would be the equivalent of `false` currently, `up` would invert the behavior and always round upwards, and `closest` would round towards the closest integer, rather than specifically up or down.

**Describe alternatives you've considered**
With the existing options the only alternatives I've come up with are calculating the padding based on the kind of rounding I want based on the difference between the date for comparison and now, or altering the date for comparison itself to generate the term I want when handed to `toRelative`. An alternative to the addition I suggest above might be to allow handing a function to the `round` option that would receive a parameter of the unrounded relative value, and the resulting number would be whatever is returned from that function.

NfNitLoop:
This "rounding down" behavior is especially strange when using `toRelative()` to show a countdown timer.

With the default behavior, you'll see "1 minute" for a full 60 seconds before you see "59 seconds" and realize *now* you've actually got a minute left.

I was hopeful when I saw https://github.com/moment/luxon/pull/926 because I thought specifying "multiple units" like `["minutes", "seconds"]` would allow me to see output like `in 1 minute 23 seconds`, but that's not how that behaves. (It continues to pick a single unit, from among those choices.)

So maybe an option like `significantUnits` would help here.  (Please feel free to tell me to open a separate issue for this if you'd like to consider it separately. I'm mostly adding it to this ticket for some example use cases.) To take the above 29.99 minutes as an example, I'd expect `.toRelative({significantUnits: 2})` to output `in 29 minutes 59 seconds`.

NfNitLoop:
I ended up implementing my own. If someone wants to fold that into Luxon as a feature, I'll be happy to use it from there. :) 

```typescript
function relativeDuration(duration: Duration, opts?: Opts): string {
    let sigU = opts?.significantUnits ?? 2
    if (sigU < 1) {
[...]

PR #1683, opened 2025-02-14
feat: more rounding options to toRelative

hugohutri:
This PR addresses the issue described in #1129 by introducing an additional rounding option: "nearest".

## Problem
Currently, the amount is always rounded down. For example, if the date is 1.99 days away, it is displayed as:

"in 1 day" (when rounded down)
"in 1.99 days" (when rounding is disabled)

## Solution
With the new "nearest" rounding option, the display will round to the closest whole number instead of always rounding down. This provides a more intuitive and accurate representation of time differences.


diesieben07:
Hi @hugohutri and thank you for the contribution. 
However I have decided to go with the more comprehensive approach of #1685

PR #1685, opened 2025-02-19
Rounding options to toRelative

### LINKED ISSUE #1129: More control over DateTime#toRelative rounding behavior

Shiva127:
This PR fully address the issue described in #1129 (and fully replace PR #1683) by introducing an additional `rounding` option. This PR is fully backward compatible.

## Problem

Before this PR, the only option for rounding with `toRelative()` is to rounds towards zero.

## Solution

With this PR, I add a new parameter `rounding` who can be `"trunc"` (towards zero), `"expand"` (away from zero), `"round"`, `"floor"`, or `"ceil"` to address all options. The `rounding` option works with `round: true` and `round: false`.

I'm open for better names for `"towardsZero"`, `"awayFromZero"`. 

Fixes #1129.

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -447,7 +447,7 @@ function quickDT(obj, opts) {
 
 function diffRelative(start, end, opts) {
   const round = isUndefined(opts.round) ? true : opts.round,
-    rounding = isUndefined(opts.rounding) ? "towardsZero" : opts.rounding,
+    rounding = isUndefined(opts.rounding) ? "trunc" : opts.rounding,
     format = (c, unit) => {
       c = roundTo(c, round || opts.calendary ? 0 : 2, opts.calendary ? "round" : rounding);
       const formatter = end.loc.clone(opts).relFormatter(opts);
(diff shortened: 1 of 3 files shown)


hugohutri:
Nice! Good that you implemented the remaining rounding methods as well 😎 
`"towardsZero"` and `"awayFromZero"` are pretty descriptive and I don't know if there are good alternatives. If we really want some other names for those, then `"towardsZero"` could be `"truncate"` or `"clip"` but awayFromZero is a bit harder to name imo 🤔 Maybe `"expand"` ?
But looks good anyway!

Shiva127:
> Nice! Good that you implemented the remaining rounding methods as well 😎

Thank you!

> `"towardsZero"` and `"awayFromZero"` are pretty descriptive and I don't know if there are good alternatives.

Yes, it's descriptive but I don't know if there is a more accurate mathematical name for that.

> If we really want some other names for those, then `"towardsZero"` could be `"truncate"` or `"clip"` but awayFromZero is a bit harder to name imo 🤔 Maybe `"expand"` ? But looks good anyway!

Shiva127:
I've found the [rounding modes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#roundingmode) of [Intl.NumberFormat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) function and they use `"expand"` for `"awayFromZero"` and `"trunc"` for `"towardsZero"`.

@hugohutri You were right. 😉

PR #1701, opened 2025-04-11
Only add minus sign on largest unit of negative durations

dobon:
Per discussion https://github.com/moment/luxon/discussions/1699#discussioncomment-12781313

### DIFF
--- a/src/impl/formatter.js
+++ b/src/impl/formatter.js
@@ -394,10 +394,10 @@ export default class Formatter {
             return null;
         }
       },
-      tokenToString = (lildur, $) => (token) => {
+      tokenToString = (lildur, info) => (token) => {
         const mapped = tokenToField(token);
         if (mapped) {
-          const inversionFactor = $.isNegativeDuration && mapped !== $.largestUnit ? -1 : 1;
+          const inversionFactor = info.isNegativeDuration && mapped !== info.largestUnit ? -1 : 1;
           return this.num(lildur.get(mapped) * inversionFactor, token.length);
         } else {
           return token;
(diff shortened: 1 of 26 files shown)


src/impl/formatter.js
```diff
@@ -375,29 +375,30 @@ export default class Formatter {
...
+            return "years";
           default:
             return null;
         }
       },
       tokenToString = (lildur) => (token) => {
         const mapped = tokenToField(token);
         if (mapped) {
-          return this.num(lildur.get(mapped), token.length);
+          const inversionFactor = lildur < 0 && mapped !== Object.keys(lildur.values)[0] ? -1 : 1;
```
diesieben07, line 400: 1. We should not rely on the order of keys in `lildur.values`
2. `lildur < 0` converts the duration to milliseconds every time. This should be done just once.

src/impl/formatter.js
```diff
@@ -375,29 +375,30 @@ export default class Formatter {
...
+            return "days";
           case "w":
-            return "week";
+            return "weeks";
           case "M":
-            return "month";
+            return "months";
           case "y":
-            return "year";
+            return "years";
```
diesieben07, line 392: Why were these changed?

diesieben07 (changes requested):

src/impl/formatter.js
dobon: So that `mapped` would match the `Object.keys` of `lildur.values`, and because `lildur.get` doesn't care whether it is plural or not.

src/impl/formatter.js
diesieben07: Not sure if there is an _easy_ way. We'd have to find the index in `orderedUnits`. `shiftTo` happens here every time, can you check if that has a defined key order? But even if it does, it feels wrong to rely on that.

dobon: [duration.js line 794](https://github.com/dobon/luxon/blob/a6c5577966f5f071343f0fe35da82b64c09189a4/src/duration.js#L794C5-L821C6)

`shiftTo` builds up `values` by looping through `orderedUnits`.

diesieben07:
Thank you!

Issue #1655, opened 2024-09-21
Parsing date from format `hh` gives valid true even if the hour is greater than 12


rohitkhatri:
**Describe the bug**
Parsing time from string based on a `hh:mm a` format gives valid true even if the hour is greater than 12

**To Reproduce**
```typescript
import { DateTime } from "luxon";

const date = DateTime.fromFormat('13:00 AM', 'hh:mm a', {});
console.log(date);
console.log(date.isValid);
```

**Actual vs Expected behavior**
It should give invalid since the definition of `hh` is 12 hours clock

**Desktop (please complete the following information):**

- OS: MacOS M1
- Browser Chrome: 123
- Luxon version: 3.1.0
- Your timezone: Asia/Kolkata

rohitkhatri:
Is there any workaround for this for now?

dobon:
Kind of a late reply, but I think this is actually expected behavior. I checked how Moment handles this situation and it also considers "13:00 AM" to be valid (and representing "1:00 PM" or "13:00 PM" or just "13:00"). Basically the rule seems to be "if a 24-hour time is provided, trust that and ignore the meridiem whether it was provider or not.
```javascript
>moment('13:00 AM', 'hh:mm a').isValid();
>>true
>moment('13:00 AM', 'hh:mm a').format('h:mm a');
>>"1:00 PM"
>moment('13:00 PM', 'hh:mm a').format('h:mm a');
>>"1:00 PM"
```

practically speaking, if this is considered invalid for your application, then you'll need to add some additional logic to do that additional validation.

diesieben07:
I agree that this is expected behavior. The general robustness principle of "be conservative in what you send, be liberal in what you accept" applies. 
However I would be open to a pull request which adds a `strict` option to `fromFormat` which enables stricter parsing.

PR #1702, opened 2025-04-14
Add strict hour option for parsing 12-hour dates

dobon:
Addressing https://github.com/moment/luxon/issues/1625
and related
https://github.com/moment/luxon/issues/1655

Adds a strict option to fromFormat which enables stricter parsing of 12-hour time formats, such that an error is thrown if the parsed hour is outside of the range 1-12.

By default strict parsing is disabled.  `DateTime.fromFormat("18:30", "h:mm", { strictHours: true })` will throw `ConflictingSpecificationError` because `strictHours` is enabled.

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -455,6 +455,15 @@ export class TokenParser {
         [result, zone, specificOffset] = matches
           ? dateTimeFromMatches(matches)
           : [null, null, undefined];
+      if (
+        this.locale.strictHours &&
+        hasOwnProperty(matches, "h") &&
+        (matches.h > 12 || matches.h < 1)
+      ) {
+        throw new ConflictingSpecificationError(
+          "Can't go over 12 or under 1 hours when specifying 12-hour format and strict hour parsing enabled"
+        );
+      }
       if (hasOwnProperty(matches, "a") && hasOwnProperty(matches, "H")) {
         throw new ConflictingSpecificationError(
           "Can't include meridiem when specifying 24-hour format"
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -447,7 +447,7 @@ export class TokenParser {
     }
   }
 
-  explainFromTokens(input) {
+  explainFromTokens(input, strictHours = false) {
     if (!this.isValid) {
       return { input, tokens: this.tokens, invalidReason: this.invalidReason };
     } else {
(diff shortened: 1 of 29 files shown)


dobon:
This duplicates (and adds the requested `strictHours` to) https://github.com/moment/luxon/pull/1631

src/impl/tokenParser.js
```diff
@@ -447,14 +447,19 @@ export class TokenParser {
...
+  explainFromTokens(input, strictHours = false) {
     if (!this.isValid) {
       return { input, tokens: this.tokens, invalidReason: this.invalidReason };
     } else {
       const [rawMatches, matches] = match(input, this.regex, this.handlers),
         [result, zone, specificOffset] = matches
           ? dateTimeFromMatches(matches)
           : [null, null, undefined];
+      if (strictHours && hasOwnProperty(matches, "h") && (matches.h > 12 || matches.h < 1)) {
+        throw new ConflictingSpecificationError(
```
diesieben07, line 459: It should produce an invalid `DateTime`, just like `DateTime.fromFormat('99:23', 'hh:mm')` does currently.
`ConflictingSpecificationError` means the programmer misused `fromFormat`.

test/datetime/tokenParse.test.js
```diff
@@ -74,6 +74,17 @@ test("DateTime.fromFormat() throws if you specify meridiem with 24-hour time", (
   expect(() => DateTime.fromFormat("930PM", "Hmma")).toThrow(ConflictingSpecificationError);
 });
 
+test("DateTime.fromFormat() throws if h is used with 24-hour time and strictHours option is enabled", () => {
+  expect(() => DateTime.fromFormat("18:30", "h:mm")).not.toThrow(ConflictingSpecificationError);
```
diesieben07, line 78: Please move this to a separate test, named something like 'h token accepts 24 hours when strictHours is disabled' and test that it is parsed properly instead of just testing that it does not throw.

test/datetime/tokenParse.test.js
```diff
@@ -74,6 +74,17 @@ test("DateTime.fromFormat() throws if you specify meridiem with 24-hour time", (
...
+  expect(() => DateTime.fromFormat("18:30", "h:mm", { strictHours: true })).toThrow(
+    ConflictingSpecificationError
+  );
+  expect(() => DateTime.fromFormat("00:30", "h:mm")).not.toThrow(ConflictingSpecificationError);
```
diesieben07, line 82: Same as above.

diesieben07 (changes requested):
What happens if the users uses a macro token such as `t` and we get `hourCycle: h11` from the options? Then we turn it into `h`, but that can't parse `0` (which is valid for h11).

test/datetime/tokenParse.test.js
```diff
@@ -165,6 +165,38 @@ test("DateTime.fromFormat() parses hours", () => {
...
+  rejects("60", "m");
+
+  accepts("59", "s");
+  rejects("60", "s");
+
+  accepts("0", "h", { strictHours: true, locale: "en-US-u-hc-h11" });
+  rejects("0", "h", { strictHours: true });
+  rejects("12", "h", { strictHours: true, locale: "en-US-u-hc-h11" });
+  accepts("12", "h", { strictHours: true });
+  rejects("13", "h", { strictHours: true });
```
diesieben07, line 197: Can you add some tests for the other hour cycles as well?

diesieben07 (changes requested):
I really apologize for taking so long to get to this and thank you again for working on it.
I suggested some improvements.

diesieben07:
Superseded by #1787

PR #1707, opened 2025-05-21
build: use the es6 build for ESM exports

pataar:
Currently it seems that Metro (from React Native) isn't resolving Luxon correctly due to the incorrect exports configuration. 

It seems that ESM imports prefer the source code instead of the built code. Changing this to the es6 build resolves the issue.

This is the error with the old export: 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "repository": "https://github.com/moment/luxon",
   "exports": {
     ".": {
-      "import": "./src/luxon.js",
+      "import": "./build/es6/luxon.js",
       "require": "./build/node/luxon.js"
     },
     "./package.json": "./package.json"


diesieben07:
This has caused problems, so I have reverted it for now until we can figure out what the problem is.
The whole build setup needs an overhaul, which I will tackle soon.

veliseev93:
Any updates?

pierpo:
Hi! I am using 3.7.1 and still have the issue on React Native (Expo 53) 🤔 It looks like the revert was not enough?

Still, I think the real issue here would be to tackle the circular dependency issue itself.
```
src/impl/locale.js -> src/settings.js -> src/impl/locale.js
```


Also, here's the error as text so that other people can find it easily (I was not able to find the issue because the error is only in the screenshot 😁 )

```
isUndefined is not a function (it is undefined)
```

mzaien:
> luxon@3.8.0-alpha.1

@diesieben07 just tried it, the release fixes it 🚀

diesieben07:
Thank you for confirming!

Issue #1726, opened 2025-08-26
Wrong key in regexParser?


tonysamperi:
**Describe the bug**
I think there's a mistake in your recent commit e434a47bc466b83c2c29766fef1feb1e7ee14a2f in regexParser.js

```ts
const extractPartialIsoIntervalEndDate = simpleParse(
  "month",
  "day",
  "weekNumber",
  "weekDay", // This is wrong
  "ordinal"
);
```

Across the whole lib, it's `weekday` not `weekDay`. Not sure this causes problems, but if you were using typings, you would get an error there.

PR #1743, opened 2025-10-13
Fix weekday naming inconsistency in regexParser

sheromero:
Addressing #1726

Renamed instances of weekDay to weekday in regexParse.js for consistency.


diesieben07:
Thank you and sorry for taking so long to get to it!

PR #1756, opened 2026-01-29
feat: add weekSettings support to DateTime.reconfigure()

### LINKED ISSUE #1746: DateTime.reconfigure does not accept weekSettings
**Describe the bug**
DateTime, Duration (and possibly others) do not accept weekSettings in .reconfigure method, even though the cloning of the locale inside can handle it. 

@types/luxon has a wrong definition or a mismatch with the description of the method.

**To Reproduce**

`Settings.defaultWeekSettings = {
  firstDay: 4,
  minimalDays: 1,
  weekend: [],
};`
`let firstDate = DateTime.now();`
`console.log(firstDate.startOf('week', { useLocaleWeeks: true }));`
`let b = firstDate.reconfigure({ weekSettings: { firstDay: 6, minimalDays: 1, weekend: [1, 2]} });`
`console.log(b.startOf('week', { useLocaleWeeks: true }));`

Actual: Both dates return the same start day.
Expected: Either typings should be updated to reflect that weedSettings are not supported or (my preferred solution ) reconfigure method should accept weekSettings and pass them to Location.clone().

- OS: Windows 11
- Browser: Chrme Version 142.0.7444.163 (Official Build) (64-bit)
- Luxon version: 3.7.2
- Your timezone: Europe/Sofia

**Additional context**
So we have this product where people are working in all parts of the world and need to collab on the same things. Given that we have a lot of personal settings related to Timezone/Start of week/Locale that we override the user`s machine settings with. I am currently doing a migration to Luxon from MomentJS (apparently Luxon deals best with timezones and locales from all other non-moment libraries).

So long story short I have a case where I need to recalculate some date ranges and some days when a user changes their start of week setting and noticed the above behavior. 

I looked at the code and noticed that if the parameter gets added to reconfigure it should work fine downstream when cloning the locale. 

Im willing to do a PR to add weekSettings as a parameter, but I wasn't sure if it would be accepted.

taronsung:
## Summary

Adds `weekSettings` support to `DateTime.reconfigure()`, allowing users to change week settings on an existing DateTime instance.

## Problem

As reported in #1746, `DateTime.reconfigure()` did not accept `weekSettings` even though:
1. The underlying `Locale.clone()` method already supports it
2. The TypeScript types suggest it should be supported
3. Other DateTime factory methods accept `weekSettings`

This made it impossible to change week settings on an existing DateTime without recreating it entirely.

## Solution

Added `weekSettings` to the destructured parameters of `reconfigure()` and pass it through to `Locale.clone()`.

## Example

```javascript
// Set custom week settings on an existing DateTime
const dt = DateTime.local(2022, 1, 4, { locale: "en-US" });
const reconfigured = dt.reconfigure({ 
  weekSettings: { firstDay: 6, minimalDays: 1, weekend: [1, 2] } 
});
reconfigured.startOf("week", { useLocaleWeeks: true }).weekday; // => 6
```

## Tests

Added tests to verify:
1. `reconfigure()` correctly sets weekSettings
2. `reconfigure()` preserves existing weekSettings when setting other options

Fixes #1746

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1581,13 +1581,21 @@ export default class DateTime {
   }
 
   /**
-   * "Set" the locale, numberingSystem, or outputCalendar. Returns a newly-constructed DateTime.
+   * "Set" the locale, numberingSystem, outputCalendar, or weekSettings. Returns a newly-constructed DateTime.
    * @param {Object} properties - the properties to set
+   * @param {string} [properties.locale] - the locale to set
+   * @param {string} [properties.numberingSystem] - the numbering system to set
+   * @param {string} [properties.outputCalendar] - the output calendar to set
+   * @param {Object} [properties.weekSettings] - the week settings to set
+   * @param {number} [properties.weekSettings.firstDay] - the first day of the week (1-7, Monday-Sunday)
+   * @param {number} [properties.weekSettings.minimalDays] - the minimum number of days in the first week
+   * @param {number[]} [properties.weekSettings.weekend] - the weekend days
    * @example DateTime.local(2017, 5, 25).reconfigure({ locale: 'en-GB' })
+   * @example DateTime.local(2017, 5, 25).reconfigure({ weekSettings: { firstDay: 1 } })
    * @return {DateTime}
    */
-  reconfigure({ locale, numberingSystem, outputCalendar } = {}) {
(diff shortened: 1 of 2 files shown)


diesieben07:
Thank you!

PR #1763, opened 2026-02-18
Render dark mode toggle in sidebar on all pages (dobon's review)

dobon:
@camoch23 I reviewed your code and it looks pretty good, but I noticed that there is already a "is page scrolled?" feature built into docsify: when you are on the cover page the body element does NOT have a 'sticky' class, and when you are in the docs the body does have a 'sticky' class active.

Because of this I was able to eliminate the IntersectionObserver (which is a cool API that I haven't used before, thanks for introducing it to me), and instead handle the fade-in/fade-out styling in CSS.

For full clarity, my commit duplicates the dark mode toggle so it appears in both the cover (when it exists) and on the sidebar, but the sidebar toggle is opacity:0.0 until body has the 'sticky' class applied (ie: when you have scrolled down past the cover page).

I didn't really want to open a new pull request, but I couldn't figure out how to submit my commit to your branch, so please forgive the duplication.

### DIFF
--- a/site/plugins/dark-theme-toggle.css
+++ b/site/plugins/dark-theme-toggle.css
@@ -18,13 +18,13 @@
   margin-top: 0;
   right: 0;
   left: unset;
-  transition: opacity 250ms ease;
-  opacity: 0.0;
-  pointer-events: none;
+  transition: opacity 250ms ease, visibility 250ms steps(1, jump-both);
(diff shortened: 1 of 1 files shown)


diesieben07:
Thank you, I do like this better than the IntersectionObserver version!

Issue #1757, opened 2026-01-30
fromFormat and unix epoch


nlooije:
**Describe the bug**
I cannot use the token 'x' and 'X' to convert a unix epoch time in (milli)seconds to a datetime using the .fromFormat() method.

**To Reproduce**
`luxon.DateTime.fromFormat('1769758221', 'X')`

**Actual vs Expected behavior**
fromFormat() method gives a unparseable error with explanation _the input "1769758221" can't be parsed as format X_ .
I expect the result to be datetime object which when converted toISO() is _2026-01-30T08:30:21.000+01:00_

**Desktop:**

- OS: Windows 11
- Browser Chrome 144.0.7559.97
- Luxon version 3.7.2
- Europe/Amsterdam

**Additional context**
I may be misunderstanding if fromFormat() method can convert from unix epoch timestamps at all. It is not clear from the documentation. There is a macro token 'x' and 'X' but it seems to work for the toFormat() method instead.

diesieben07:
@JacquesN16 Sure! If this is a feature you need or you just want to help out, feel free to send in a PR.

PR #1769, opened 2026-03-23
fix: support X and x tokens for unix epoch parsing

JacquesN16:
**Fix issue** [#1757 ](https://github.com/moment/luxon/issues/1757)
The DateTime.fromFormat not supporting X and x unix timestamp tokens
Adds X (unix seconds) and x (unix milliseconds) token support to _fromFormat_ by extending _unitForToken_ .
Also, handles edge-case for epoch zero and negative timestamps (pre-1970).
Before:
  DateTime.fromFormat("1769758221", "X")  // Invalid DateTime 
  DateTime.fromFormat("1769758221000", "x")  // Invalid DateTime
After:
  DateTime.fromFormat("1769758221", "X").toUTC().toISO()   // "2026-01-30T07:30:21.000Z"
  DateTime.fromFormat("1769758221000", "x").toUTC().toISO() // "2026-01-30T07:30:21.000Z"

### DIFF
--- a/src/datetime.js
+++ b/src/datetime.js
@@ -1001,9 +1001,15 @@ export default class DateTime {
         numberingSystem,
         defaultToEN: true,
       }),
-      [vals, parsedZone, specificOffset, invalid] = parseFromTokens(localeToUse, text, fmt);
+      [vals, parsedZone, specificOffset, invalid, knownEpochMs] = parseFromTokens(
+        localeToUse,
+        text,
+        fmt
+      );
     if (invalid) {
       return DateTime.invalid(invalid);
+    } else if (knownEpochMs !== undefined) {
+      return DateTime.fromMillis(knownEpochMs, opts);
     } else {
       return parseDataToDateTime(vals, parsedZone, opts, `format ${fmt}`, text, specificOffset);
     }
@@ -2429,10 +2435,13 @@ export default class DateTime {
       );
     }
 
-    const { result, zone, specificOffset, invalidReason } = formatParser.explainFromTokens(text);
+    const { result, zone, specificOffset, invalidReason, knownEpochMs } =
+      formatParser.explainFromTokens(text);
 
     if (invalidReason) {
       return DateTime.invalid(invalidReason);
+    } else if (knownEpochMs !== undefined) {
+      return DateTime.fromMillis(knownEpochMs, opts);
     } else {
       return parseDataToDateTime(
         result,
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -5,6 +5,7 @@ import IANAZone from "../zones/IANAZone.js";
 import DateTime from "../datetime.js";
 import { digitRegex, parseDigits } from "./digits.js";
 import { ConflictingSpecificationError } from "../errors.js";
+import Invalid from "./invalid.js";
 
 const MISSING_FTP = "missing Intl.DateTimeFormat.formatToParts support";
 
@@ -470,6 +471,20 @@ export class TokenParser {
           "Can't include meridiem when specifying 24-hour format"
         );
       }
+      if (hasOwnProperty(matches, "X") && hasOwnProperty(matches, "x")) {
+        if (matches.X * 1000 !== matches.x) {
+          return {
+            input,
+            tokens: this.tokens,
+            invalidReason: new Invalid(
+              "mismatched unix timestamp",
+              `X=${matches.X} (${matches.X * 1000}ms) and x=${
+                matches.x
+              } specify different points in time`
+            ),
+          };
+        }
+      }
       if (knownEpochMs !== undefined && result && Object.keys(result).length > 0) {
         // X is unix seconds with no sub-second precision; S/SSS/u add the missing ms part
         if (hasOwnProperty(matches, "X") && result.millisecond !== undefined) {
@@ -481,9 +496,14 @@ export class TokenParser {
           const epochDateTime = DateTime.fromMillis(knownEpochMs, { zone: zone ?? "UTC" });
           for (const [key, value] of Object.entries(result)) {
             if (epochDateTime[key] !== value) {
-              throw new ConflictingSpecificationError(
-                `Can't specify ${key} as ${value} when the unix timestamp implies ${epochDateTime[key]}`
-              );
+              return {
+                input,
+                tokens: this.tokens,
+                invalidReason: new Invalid(
+                  `mismatched ${key}`,
+                  `you can't specify ${key}=${value} when the unix timestamp implies ${key}=${epochDateTime[key]}`
+                ),
+              };
             }
           }
         }
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -470,15 +470,9 @@ export class TokenParser {
         );
       }
       if (knownEpochMs !== undefined && result && Object.keys(result).length > 0) {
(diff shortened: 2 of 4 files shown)


test/datetime/tokenParse.test.js
```diff
@@ -1248,3 +1248,52 @@ test("DateTime.fromFormatParser throws error when used with a different locale t
...
+  expect(dt.minute).toBe(59);
+  expect(dt.second).toBe(59);
+});
+
+test("fromFormat with x for negative unix milliseconds (before 1970)", () => {
+  const dt = DateTime.fromFormat("-1000", "x", { zone: "UTC" });
+  expect(dt.isValid).toBe(true);
+  expect(dt.valueOf()).toBe(-1000);
+  expect(dt.year).toBe(1969);
+});
```
diesieben07, line 1299: Please add some "negative" tests as well, i.e. tests that ensure it doesn't erroneously parse decimal numbers or other nonsense.

src/datetime.js
```diff
@@ -2429,10 +2435,13 @@ export default class DateTime {
...
     }
 
-    const { result, zone, specificOffset, invalidReason } = formatParser.explainFromTokens(text);
+    const { result, zone, specificOffset, invalidReason, knownEpochMs } =
+      formatParser.explainFromTokens(text);
 
     if (invalidReason) {
       return DateTime.invalid(invalidReason);
+    } else if (knownEpochMs !== undefined) {
+      return DateTime.fromMillis(knownEpochMs, opts);
```
diesieben07, line 2444: Currently when `x` or `X` is present in the format it overrules anything else. It needs to throw `ConflictingSpecificationError` if there are other tokens in the format and the parsed string is not consistent, i.e. the unix timestamp doesn't match what the other tokens say.
You can probably do this best in `TokenParser#explainFromTokens` where another case like this is already handled.
Most of this kind of logic is already provided by DateTime#fromObject, which is what ultimately gets fed the object, but it does not accept (and therefor does not handle) all things that can be parsed.

src/impl/tokenParser.js
```diff
@@ -181,6 +181,11 @@ function unitForToken(token, loc) {
         // because we don't have any way to figure out what they are
         case "z":
           return simple(/[a-z_+-/]{1,256}?/i);
+        // unix timestamps
+        case "X":
+          return intUnit(/[+-]?\d+/);
+        case "x":
+          return intUnit(/[+-]?\d+/);
```
diesieben07, line 188: Please update the "Table of tokens" in `docs/parsing.md` with the new tokens.

diesieben07 (changes requested):
This is a good start, thank you for the contribution! 
Some small things are still missing.

src/impl/tokenParser.js
JacquesN16: I have update the table of tokens with new tokens and the output examples.

src/datetime.js
JacquesN16: I have made a simple fix.
Mixing X/x with any other format token now always throws a ConflictingSpecificationError, regardless of consistency. This should help prevent token misuse. 
I don’t see a clear use case for supporting syntax like DateTime.fromFormat("2026 1769758221000", "yyyy x").
What do you think?

test/datetime/tokenParse.test.js
JacquesN16: @diesieben07 thanks for your feedback. 
I have add new tests to cover those cases.

src/datetime.js
diesieben07: Really sorry for the long silence. Unfortunately I don't think this is really sufficient.
At the very least tokens related to milliseconds / fractional seconds need to be supported as well. And this would be a clear deviation from all other tokens. If you include day, month, year _and_ weekday, that is technically redundant, but the parser will allow it and validate that the parsed data is consistent.

JacquesN16: Thanks for the feedback. 
 I've updated the implementation to align with how Luxon already handles redundant tokens:
- when X/x is combined with other format tokens, the parser now validates consistency rather
than always throwing. So DateTime.fromFormat("2026 1769758221", "yyyy X") succeeds (year
matches the epoch), while DateTime.fromFormat("1970 1769758221", "yyyy X") throws a
ConflictingSpecificationError.

- For milliseconds/fractional seconds alongside X, since X carries no
sub-second precision, S/SSS/u are treated as additive rather than redundant. 
So, DateTime.fromFormat("1769758221 500", "X SSS") produces epoch 1769758221500ms. For x (unix
milliseconds), millisecond tokens follow the same consistency-check path as all other
tokens.

  Let me know if I'm still missing something or if you'd like any adjustments.

diesieben07: Thanks, this does what I wanted know, but it turns out _I_ was wrong.
For some reason I had in mind that if this happens we throw ConflictingSpecificationError, but it should really produce an invalid DateTime.
 If you look at DateTime.fromObject at the very end is where it handles this case (for weekdays, i.e. there is a full date in the format, but also the weekday; look for `"mismatched weekday"`).

When this happens, this not really the programmer's fault (which would be indicated by ConflictingSpecificationError), but the parsed data is invalid.
This is on me and I gave you wrong guidance. If you're up for changing it I would appreciate that, otherwise I can fix it up as well, since I asked for the wrong thing. Apologies!

Edit: Oh, and we also need to handle the case where x _and_ X are in the format string.

JacquesN16: No worries at all and thanks for the clarification! 
The distinction makes sense. I've updated the implementation to follow the same pattern as the "mismatched weekday" case in fromObject: when a redundant token is inconsistent with the epoch, we now return DateTime.invalid("mismatched <field>", ...) rather than throwing.
I've handled case when both X and x are present, the parser now validates that they point to the same instant. Consistent values succeed, inconsistent ones produce DateTime.invalid().
Tests were also added to cover these cases.

src/impl/tokenParser.js
```diff
@@ -449,14 +462,52 @@ export class TokenParser {
...
+      }
+      if (knownEpochMs !== undefined && result && Object.keys(result).length > 0) {
+        // X is unix seconds with no sub-second precision; S/SSS/u add the missing ms part
+        if (hasOwnProperty(matches, "X") && result.millisecond !== undefined) {
+          knownEpochMs = knownEpochMs + result.millisecond;
+          delete result.millisecond;
+        }
+        // All remaining tokens are redundant — validate they are consistent with the epoch
+        if (Object.keys(result).length > 0) {
+          const epochDateTime = DateTime.fromMillis(knownEpochMs, { zone: zone ?? "UTC" });
```
diesieben07, line 496: Defaulting the Zone to UTC is not correct for two reasons:
1. `fromFormat` accepts a `zone` option, which has to be honored
2. If `zone` is not set via `fromFormat` options _and_ there is no zone in the parsed data then the default is `Settings.defaultZone`, not UTC.

For this reason it might be easier to move this logic to `parseDataToDateTime` as well (see the other comment), because that is where all the information comes together and you know the final zone.

src/datetime.js
diesieben07, line 2444: This will not handle the `setZone` option. This logic of handling `knownEpochMs` should really be in `parseDataToDateTime` (where `setZone` is handled), then you don't have to duplicate it between `fromFormat` and `fromFormatParser` either.

diesieben07 (changes requested):

PR #1775, opened 2026-04-24
Fix: wrong argument passed to parseDataToDateTime in fromHTTP

bravesasha:
The fifth argument to parseDataToDateTime should be the original input text used to build the error message on parse failure. Instead, opts was passed, causing unparsable HTTP date errors to report object Object as the input string instead of the actual value.


diesieben07:
Thank you!

PR #1781, opened 2026-06-10
Fix FixedOffsetZone reporting isValid=true for unsupported inputs

### LINKED ISSUE #1068: Setting FixedOffsetZone as option.zone for DateTime.fromISO does not work
**Describe the bug**
When I set timezone with a ISO string as 'America/Chicago', the DateTime was created properly.
When I set timezone with a FixedOffsetZone object, the DateTime was created invalid.

**To Reproduce**
Please share a minimal code example that triggers the problem:
```
> var zone = new luxon.FixedOffsetZone('CDT');
undefined
> zone.isValid
true
> var d1 = luxon.DateTime.fromISO('2021-11-03T13:00:00', { zone: 'America/Chicago'});
undefined
> d1
DateTime {
  ts: 1635962400000,
  _zone: IANAZone { zoneName: 'America/Chicago', valid: true },
  loc: Locale {
    locale: 'en-US',
    numberingSystem: null,
    outputCalendar: null,
    intl: 'en-US',
    weekdaysCache: { format: {}, standalone: {} },
    monthsCache: { format: {}, standalone: {} },
    meridiemCache: null,
    eraCache: {},
    specifiedLocale: null,
    fastNumbersCached: null
  },
  invalid: null,
  weekData: null,
  c: {
    year: 2021,
    month: 11,
    day: 3,
    hour: 13,
    minute: 0,
    second: 0,
    millisecond: 0
  },
  o: -300,
  isLuxonDateTime: true
}
> var d2 = luxon.DateTime.fromISO('2021-11-03T13:00:00', { zone: zone });
undefined
> d2
DateTime {
  ts: NaN,
  _zone: FixedOffsetZone { fixed: 'CDT' },
  loc: Locale {
    locale: 'en-US',
    numberingSystem: null,
    outputCalendar: null,
    intl: 'en-US',
    weekdaysCache: { format: {}, standalone: {} },
    monthsCache: { format: {}, standalone: {} },
    meridiemCache: null,
    eraCache: {},
    specifiedLocale: null,
    fastNumbersCached: null
  },
  invalid: Invalid { reason: 'invalid input', explanation: undefined },
  weekData: null,
  c: null,
  o: null,
  isLuxonDateTime: true
}
```

I expect that d2 is valid because the timzone is valid and CDT and America/Chicago is basically the same.

- Browser [e.g. Chrome 84, safari 14.0]
- Luxon version [e.g. 1.25.0]
- Your timezone [e.g. "America/New_York"]

**Additional context**

chatman-media:
Closes #1068

## The bug

`FixedOffsetZone` is a public, exported class, but its constructor performs no validation and `isValid` unconditionally returns `true`. As a result a zone built from an input it does not support still claims to be valid:

```js
const zone = new luxon.FixedOffsetZone("CDT");
zone.isValid;        // true  (should be false)
zone.offset(0);      // "CDT"
zone.name;           // "UTC-NaN"
zone.formatOffset(0, "short"); // "-NaN:NaN"
```

Because the zone lies about its validity, a `DateTime` that uses it slips past the existing `zone.isValid` guards in `datetime.js` and ends up invalid for the wrong reason (`"invalid input"` downstream) rather than being cleanly rejected.

This was confirmed as a bug by the maintainer in the issue:

> the bug here is not that the zone isn't working; it's that `isValid` should return false. "CDT" is not a valid input for building a zone.

`FixedOffsetZone` exposes factory methods (`parseSpecifier`, `instance`) that already return `null`/numbers for bad input, but the public constructor was never protected, so this only manifests when the class is instantiated directly.

## The fix

The constructor now records whether it was given a numeric offset, and `isValid` returns that flag:

```js
this.valid = Number.isInteger(offset);
```

- Valid numeric offsets (incl. `0` and `-0`) remain valid — no behavior change for any existing/internal caller (all internal call sites pass integers from `signedOffset(...)` or matched numeric captures).
- Non-numeric inputs (`"CDT"`, `"5"`, `NaN`, `undefined`, `null`, `{}`, …) now produce an invalid zone, mirroring how `IANAZone.create("Bogus/Zone")` already behaves.
- A `DateTime` built in such a zone is now itself invalid with the meaningful reason `"unsupported zone"`.

## Tests

Added to `test/zones/fixedOffset.test.js`:
- numeric offsets stay valid,
- non-numeric inputs are invalid,
- a `DateTime` in an invalid `FixedOffsetZone` is invalid with reason `"unsupported zone"`.

The two behavioral tests fail on `master` and pass with this change. The full zone test suite passes; the only failures in the repo (`format.test.js` / `toFormat.test.js`) are pre-existing locale/ICU-version differences unrelated to this change.

### DIFF
--- a/src/zones/fixedOffsetZone.js
+++ b/src/zones/fixedOffsetZone.js
@@ -50,6 +50,8 @@ export default class FixedOffsetZone extends Zone {
     super();
     /** @private **/
     this.fixed = offset;
+    /** @private **/
+    this.valid = Number.isInteger(offset);
   }
 
   /**
@@ -140,11 +142,11 @@ export default class FixedOffsetZone extends Zone {
 
   /**
    * Return whether this Zone is valid:
-   * All fixed offset zones are valid.
+   * All fixed offset zones are valid, provided they were constructed with a numeric offset.
    * @override
    * @type {boolean}
    */
   get isValid() {
-    return true;
+    return this.valid;
   }
 }
--- a/test/zones/fixedOffset.test.js
+++ b/test/zones/fixedOffset.test.js
@@ -1,5 +1,5 @@
 /* global test expect */
-import { FixedOffsetZone, IANAZone } from "../../src/luxon";
+import { DateTime, FixedOffsetZone, IANAZone } from "../../src/luxon";
 
 test("FixedOffsetZone.utcInstance returns a singleton", () => {
   expect(FixedOffsetZone.utcInstance).toBe(FixedOffsetZone.utcInstance);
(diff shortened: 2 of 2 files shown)


diesieben07:
Thank you for the fix!

PR #1783, opened 2026-06-17
Match any whitespace for a whitespace literal in fromFormat

### LINKED ISSUE #1619: Luxon "toLocaleString" uses an unusal space unicode when "setZone" flag is set
**Describe the bug**
When converting a time to simple time "TIME_SIMPLE" the returned string have an unusual space. such that if you happen to want to reuse it and create a new DateTime object it will not be accepted. this only happens when setzone flag is set `{ setZone: true }`

**To Reproduce**
```
  let y = DateTime.fromISO("2024-04-17T"+"10:30:00.000"+"Z", { setZone: true })
  let time = y.toLocaleString(DateTime.TIME_SIMPLE)
  let compare = "10:30 AM" == time
  console.log(compare)
  let parsedTime = DateTime.fromFormat(time, "h:mm a");

  console.log(parsedTime.toFormat("HH:mm"));
```
Actual:
Returns the string "Invalid DateTime"

Expected :
Should parse the time normally and output 10:30 

- OS: Winodws
- Browser: chrome
- Luxon version: 3.4.4

naveentehrpariya:
Fixes #1619.

### Problem

Modern ICU has `toLocaleString` place a non-breaking (U+00A0) or narrow non-breaking (U+202F) space before the meridiem. As a result, a string Luxon itself produced cannot be parsed back:

```js
const s = DateTime.fromISO("2024-04-17T10:30:00.000Z", { setZone: true })
  .toLocaleString(DateTime.TIME_SIMPLE); // "10:30 AM"

DateTime.fromFormat(s, "h:mm a").isValid; // false
```

### Cause

A whitespace literal in the format string is escaped by `escapeToken`, which turned a space into `\<space>` — a regex matching only a regular space (U+0020). The macro-token path (`case " "` → `simple(/[^\S\n\r]/)`) and the token-list path (`fixListRegex` → `\s`) already accept any whitespace; only the plain literal path did not.

### Fix

Escape whitespace in literals to `\s` so it matches any whitespace, mirroring the existing `fixListRegex` behaviour. Non-whitespace literals are unaffected (still escaped exactly).

### Tests

Added a test asserting `fromFormat("10:30<space>AM", "h:mm a")` is valid for a regular space, U+00A0, and U+202F. It fails without this change and passes with it. The existing meridiem/NBSP test (which covers NBSP *inside* the meridiem token) still passes.

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -46,7 +46,12 @@ function simple(regex) {
 }
 
 function escapeToken(value) {
-  return value.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, "\\$&");
+  // Match any whitespace where a literal contains whitespace, so that strings
+  // produced by `toLocaleString` (which can use NBSP ` ` or narrow NBSP
+  // ` ` around the meridiem) round-trip back through `fromFormat`.
+  return value.replace(/[\-\[\]{}()*+?.,\\\^$|#\s]/g, (match) =>
+    /\s/.test(match) ? "\\s" : `\\${match}`
+  );
 }
 
 /**
--- a/test/datetime/tokenParse.test.js
+++ b/test/datetime/tokenParse.test.js
@@ -109,6 +109,22 @@ test("DateTime.fromFormat() makes dots optional and handles non breakable spaces
   });
 });
 
+// `toLocaleString` can emit a non-breaking (U+00A0) or narrow non-breaking
+// (U+202F) space before the meridiem, so a whitespace literal in the format
+// should match any whitespace to allow round-tripping.
+// See https://github.com/moment/luxon/issues/1619
+test("DateTime.fromFormat() matches any whitespace for a whitespace literal", () => {
+  const nbsp = String.fromCharCode(0x00a0);
+  const narrow = String.fromCharCode(0x202f);
+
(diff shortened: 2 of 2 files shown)


diesieben07:
As [I already outlined](https://github.com/moment/luxon/issues/1619#issuecomment-2061664708) in the ticket you linked, trying to parse `toLocaleString` with a _manually constructed format_ is a fruitless endeavour.

Your change is a breaking change and cannot be accepted as-is. If the format specified by the user contains a space, a space will be matched, not arbitrary whitespace. 

To parse the output from `toLocaleString` you must use the macro tokens (like `t` for `TIME_SIMPLE`):
```ts
let dt = DateTime.fromISO("2026-04-01T13:24:56Z", { setZone: true });
let formatted = dt.toLocaleString(DateTime.TIME_SIMPLE, { locale: "en-US" });
let parsedTime = DateTime.fromFormat(formatted, "t", { locale: "en-US" });
console.log(parsedTime.toISO());
```

This will properly handle "weird whitespace" in the locale string.

I am hesitant to add more tokens to the format parsing, because the table is already huge. If you need to parse strings that contain strange white-space (I _highly_ suggest you use a stable format like ISO), you should pre-process them before passing them to `fromFormat`.

Can you present an actual use-case where you need this behavior? I am sure we can find a better solution.

Issue #1311, opened 2022-10-21
Duration.fromISO() doesn't accept some ISO string that was created by Duration.toISO()


CorentinMartinot:
**Describe the bug**

Parser `luxon.Duration.fromISO` does not accept enough decimals. 
It comes not to accept some ISO string that was created by luxon itself.

**To Reproduce**
```
const duration = luxon.Duration.fromObject({ minutes: 1 });
const durationInHours = duration.shiftTo('hours');
const durationISO = durationInHours.toISO(); // PT0.016666666666666666H

const parsedDuration1 = luxon.Duration.fromISO(durationISO);
// Error: Invalid Duration: unparsable: the input "PT0.016666666666666666H"
// can't be parsed as ISO 8601

const parsedDuration2 = luxon.Duration.fromISO('PT0.016666666H');
// with 9 decimals it works

**Actual vs Expected behavior**
I think the parser should be able to parse any JS float (wich comes with a maximum of 17 decimal I think).
Or at least, it should parse a string that was created by the lib itself.

**Desktop (please complete the following information):**

- OS: node v14.18.0
- Browser : none
- Luxon version 2.3.1
- Your timezone [e.g. "Europe/Paris"]

**Additional context**

thanks for your commitment !

PR #1784, opened 2026-06-18
fix: avoid exponential notation in Duration#toISO

spokodev:
### Problem

`Duration#toISO()` builds the string by concatenating each field value directly (`this.hours + "H"`, etc.). When a field holds a very small magnitude, JavaScript renders it in exponential notation, which is **not valid ISO 8601** and which `Duration.fromISO` cannot parse — so `toISO()` produces output its own parser rejects:

```js
const d = Duration.fromObject({ milliseconds: 1 }).shiftTo("hours");
d.toISO();                       // "PT2.7777777777777776e-7H"
Duration.fromISO(d.toISO()).isValid;   // false

Duration.fromObject({ hours: 1e-7 }).toISO();   // "PT1e-7H"  (unparsable)
Duration.fromObject({ days: 1e-7 }).toISO();    // "P1e-7D"   (unparsable)
```

Converting a small duration to a coarser unit — a normal operation — is enough to reach it.

### Fix

Route each field value through a small helper that expands exponential notation to a plain decimal, so `toISO()` always emits valid ISO 8601 that round-trips through `fromISO()`:

```js
Duration.fromObject({ hours: 1e-7 }).toISO();   // "PT0.0000001H"
Duration.fromISO("PT0.0000001H").hours;         // 1e-7
```

Values that already render without exponential notation are returned unchanged, so existing output is byte-identical (the helper is a no-op for them). Magnitudes `>= 1e21` keep exponential notation from `toFixed`, so they're untouched here — those additionally run into the parser's separate `\d{1,20}` digit cap (see #1311) and are out of scope for this change.

### Tests / verification

Added `toISO` assertions for the no-exponential output and a `fromISO(toISO())` round-trip. `test/duration/format.test.js` is green (50/50). Across the full suite the change adds 2 passing tests and introduces no new failures (verified by stash-comparing failure counts: 46 failed both with and without the change — those are the pre-existing timezone/locale environment failures that need luxon's CI `TZ=America/New_York` + full-ICU setup, unrelated to this change). `prettier --check` passes.

### DIFF
--- a/src/duration.js
+++ b/src/duration.js
@@ -209,6 +209,15 @@ function removeZeroes(vals) {
   return newVals;
 }
 
+// Render a number for toISO(). JS prints very small magnitudes in exponential
+// notation (e.g. `1e-7`), which is not valid ISO 8601 and which fromISO() cannot
+// parse, so expand those to a plain decimal. (toFixed keeps exponential notation
+// for magnitudes >= 1e21, so very large durations are left untouched here.)
+function toISONumber(value) {
+  const str = `${value}`;
+  return str.includes("e") ? value.toFixed(20).replace(/\.?0+$/, "") : str;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


diesieben07:
Sure, this is a valid edge-case. Thank you for the fix.

PR #1786, opened 2026-06-21
Fix lost negative sign in Duration#toFormat when largest unit is zero (signMode negativeLargestOnly)

chatman-media:
With `signMode: "negativeLargestOnly"`, the negative sign is shown only on the largest formatted unit. When that unit rounds to zero, `"auto"` sign display dropped the sign entirely, so a negative duration formatted identically to a positive one.

### Reproduction
```js
Duration.fromObject({ minutes: -30 }).toFormat("h:mm", { signMode: "negativeLargestOnly" })
// => "0:30"  (looks positive; expected "-0:30")
```
The largest unit (`hours`) rounds to `0`, and positive zero formats without a sign, so the whole duration reads as positive.

### Fix
In `src/impl/formatter.js`, when `signMode === "negativeLargestOnly"`, the token is the largest unit, the duration is negative, and the value is `0`, emit **negative zero** (`-0`) for that unit so the sign is preserved. Everything else is untouched: positive durations, other sign modes, and non-zero largest units behave exactly as before.

### Tests
Added a regression test in `test/duration/format.test.js`. It **fails without** the change (`"0:30"` instead of `"-0:30"`) and **passes with** it. Full `test/duration/format.test.js` suite passes (49/49).

### DIFF
--- a/src/impl/formatter.js
+++ b/src/impl/formatter.js
@@ -412,7 +412,21 @@ export default class Formatter {
             // "auto" and "negative" are the same, but "auto" has better support
             signDisplay = "auto";
           }
-          return this.num(lildur.get(mapped) * inversionFactor, token.length, signDisplay);
+          let value = lildur.get(mapped) * inversionFactor;
+          // When the negative sign is only shown on the largest unit (signMode
+          // "negativeLargestOnly") and that unit rounds to zero, "auto" sign display
+          // would drop the sign entirely (positive zero formats without a sign),
+          // making a negative duration indistinguishable from a positive one. Use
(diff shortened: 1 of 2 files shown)


diesieben07:
This really is only a partial fix, because this already behaves strange without `negativeLargestOnly`:
```js
Duration.fromObject({ hours: -10, minutes: -30 }).toFormat("h:mm") // "-10:-30"
Duration.fromObject({ minutes: -30 }).toFormat("h:mm") // "0:-30" - arguably should be "-0:-30" for consistency
```

But this narrow fix seems reasonable as a band-aid until the redesign of Duration in v4.

Thank you!

PR #1787, opened 2026-06-23
Reject out-of-range 12-hour values with a meridiem in fromFormat

### LINKED ISSUE #1625: DateTime fromFormat() is considers non-valid 24-hr formats as valid

sayantanmandal1:
Fixes #1625. `DateTime.fromFormat("18:30 AM", "h:mm a").isValid` returned `true` even though `18` is not a valid hour for the 12-hour `h` token. The parser now validates that, when a meridiem is present, the `h` value is within `[1, 12]`, surfacing an invalid result through luxon's existing invalid-reason channel. Valid inputs are unchanged. Added a regression test.

Closes #1625

### DIFF
--- a/src/impl/tokenParser.js
+++ b/src/impl/tokenParser.js
@@ -368,8 +368,15 @@ function dateTimeFromMatches(matches) {
     matches.M = (matches.q - 1) * 3 + 1;
   }
 
+  let hourInvalidReason;
+
   if (!isUndefined(matches.h)) {
-    if (matches.h < 12 && matches.a === 1) {
+    if (!isUndefined(matches.a) && (matches.h < 1 || matches.h > 12)) {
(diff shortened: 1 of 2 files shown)


diesieben07:
I like this solution of only enforcing the range on 12-hour tokens when a meridiem is present, not sure why I did not think of this before!
Thank you for the contribution!

PR #1790, opened 2026-07-13
fix(interval): respect endpoint zones in hasSame for empty intervals

### LINKED ISSUE #1424: Interval.hasSame ignores timezones unless the interval is empty
`Interval.hasSame` ignores timezones **unless the interval is empty**.

**Baseline**
`Interval.hasSame` (expectedly) returns
- `false` for `12:15 UTC+2` and `11:30 UTC+1` (different unix timestamp, same UTC hours, different local hours)
- `true ` for `12:15 UTC+2` and `12:30 UTC+2` (different unix timestamp, different UTC hours, same local hours)

because the local hours are matching.

That is because `Interval.hasSame` internally uses `DateTime.hasSame` which ignores timezones when comparing unit values.
https://github.com/moment/luxon/blob/d30908f973359e8c4127535858e48599d7e0b7c0/src/datetime.js#L1950

**Expected behaviour**
I would **expect** `Interval.hasSame` to return
- `false` for `12:15 UTC+2` and `11:15 UTC+1` (**same** unix timestamp, same UTC hours, different local hours)

because that should be the same case as the first example above.

**Actual behaviour**
Instead `Interval.hasSame` always returns `true` if the interval `isEmpty` (-> same unix timestamps) even if different timezones would result in different local unit values.

**Analysis**
This bug (?) seems to have been introduced in https://github.com/moment/luxon/issues/709 by returning `true` early in case the interval `isEmpty`.
https://github.com/moment/luxon/blob/33f79577a16f56629ddf9c1f1034259da7222d9c/src/interval.js#L251-L253

**Example code**
```js
(() => {
let s,e,i;

const log = (expect) => {
console.log(s.hour, s.toLocaleString(luxon.DateTime.DATETIME_FULL_WITH_SECONDS), s.toUTC().toISO());
console.log(e.hour, e.toLocaleString(luxon.DateTime.DATETIME_FULL_WITH_SECONDS), e.toUTC().toISO());
console.log('i.hasSame("hour")', i.hasSame("hour"), "expect", expect);
};
    
s = luxon.DateTime.fromISO("2023-01-01T10:15:00.000+00:00", {zone: 'UTC+2'});
e = luxon.DateTime.fromISO("2023-01-01T10:30:00.000+00:00", {zone: 'UTC+1'});
i = luxon.Interval.fromDateTimes(s, e);
log(false);

// 12 '1 January 2023 at 12:15:00 GMT+2' '2023-01-01T10:15:00.000Z'
// 11 '1 January 2023 at 11:30:00 GMT+1' '2023-01-01T10:30:00.000Z'
// i.hasSame("hour") false expect false
    
s = luxon.DateTime.fromISO("2023-01-01T10:15:00.000+00:00", {zone: 'UTC+2'});
e = luxon.DateTime.fromISO("2023-01-01T11:30:00.000+00:00", {zone: 'UTC+1'});
i = luxon.Interval.fromDateTimes(s, e);
log(true);

// 12 '1 January 2023 at 12:15:00 GMT+2' '2023-01-01T10:15:00.000Z'
// 12 '1 January 2023 at 12:30:00 GMT+1' '2023-01-01T11:30:00.000Z'
// i.hasSame("hour") true expect true
    
s = luxon.DateTime.fromISO("2023-01-01T10:15:00.000+00:00", {zone: 'UTC+2'});
e = luxon.DateTime.fromISO("2023-01-01T10:15:00.000+00:00", {zone: 'UTC+1'});
i = luxon.Interval.fromDateTimes(s, e);
log(false);

// 12 '1 January 2023 at 12:15:00 GMT+2' '2023-01-01T10:15:00.000Z'
// 11 '1 January 2023 at 11:15:00 GMT+1' '2023-01-01T10:15:00.000Z'
// i.hasSame("hour") true expect false
})();
```

P.S.: Thank's for providing this lib!

xianjianlf2:
## Problem

`Interval#hasSame` short-circuited to `true` for any empty interval, ignoring the case where the two endpoints share an instant but sit in different zones with differing local unit values.

## Fix

Compare the endpoints directly for empty intervals so a zone mismatch is respected:

```js
hasSame(unit) {
  if (!this.isValid) return false;
  return this.isEmpty() ? this.s.hasSame(this.e, unit) : this.e.minus(1).hasSame(this.s, unit);
}
```

## Testing

Added a test in `test/interval/info.test.js` covering an empty interval whose endpoints share an instant but sit in different zones. Existing `hasSame` tests continue to pass.

Closes #1424

### DIFF
--- a/src/interval.js
+++ b/src/interval.js
@@ -285,7 +285,10 @@ export default class Interval {
    * @return {boolean}
    */
   hasSame(unit) {
-    return this.isValid ? this.isEmpty() || this.e.minus(1).hasSame(this.s, unit) : false;
+    if (!this.isValid) return false;
+    // For an empty interval, compare the endpoints directly so that endpoints
+    // in different zones (with differing local unit values) are respected.
+    return this.isEmpty() ? this.s.hasSame(this.e, unit) : this.e.minus(1).hasSame(this.s, unit);
   }
 
   /**
--- a/test/interval/info.test.js
+++ b/test/interval/info.test.js
@@ -260,3 +260,13 @@ test.each([
     i = Interval.fromDateTimes(n, n);
   expect(i.hasSame("day")).toBe(true);
 });
+
+test("Interval#hasSame respects the zones of an empty interval's endpoints", () => {
+  // Same instant, but the endpoints are in different zones so their local
+  // hours differ (12:15 in UTC+2 vs 11:15 in UTC+1).
+  const s = DateTime.fromISO("2023-01-01T10:15:00.000+00:00", { zone: "UTC+2" }),
+    e = DateTime.fromISO("2023-01-01T10:15:00.000+00:00", { zone: "UTC+1" }),
+    i = Interval.fromDateTimes(s, e);
+  expect(i.isEmpty()).toBe(true);
+  expect(i.hasSame("hour")).toBe(false);
+});


diesieben07:
Thank you!

